PR #6, opened 2019-11-06
[EPO-2321] Basic TOC

blnkt:


src/components/site/tableOfContents/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+    icon: 'style',
+    primaryText: 'Style Guide',
+    active: false,
+  },
+  { divider: true },
+  {
+    primaryText: 'Table of Contents',
+    subheader: true,
+  },
+];
```
blnkt: Unless there's a reason this has to live outside the class, let's move these constants into the constructor like:
```
this.To_PREFIX = ....
this.routes = ....
```
and of course then update the use of these variables accordingly.

src/components/site/tableOfContents/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+@reactn
+class TableOfContents extends React.PureComponent {
+  // constructor(props) {
+  //   super(props);
+  //
+  // }
+
+  componentDidMount() {
+    const { navLinks } = this.props;
+    const links = routes.concat(navLinks);
```
blnkt: No change necessary but thought I would point out that this can also be written using the spread operator:
```
const links = [...this.routes, ...navlinks]
```
which is just kinda cool.

src/components/site/tableOfContents/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+      if (link.slug) {
+        link.component = Link;
+        link.label = link.title;
+        link.to = `${TO_PREFIX}${link.slug}`;
+        link.primaryText = link.title;
+        link.active = false;
+      }
+      return link;
+    });
+  }
```
blnkt: Instead of creating, and using/subsequently mutating the Class variable `this.navLinks`, the more Reacty way to accomplish this is actually to encapsulate all of this routes/navLinks logic into a class method that returns the navLinks, and then call that method  when you're setting the `navItems` prop on line 81.  This will actually also set you up to be updating that active state stuff in one place as well (since the render method runs on both mount and update).

src/components/site/tableOfContents/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+    const { TEMPORARY } = Drawer.DrawerTypes;
+    const { visible } = this.props;
+
+    return (
+      <Drawer
+        type={TEMPORARY}
+        className={styles.tableOfContents}
+        visible={visible}
+        onVisibilityChange={this.handleVisibility}
+        navItems={this.navLinks}
```
blnkt: per the comment above, this would now look like `navItems={this.getNavLinks(this.props.navLinks)}`.  Or something similar.  In react world, as much as possible, if I'm writing these kinds of helper methods, I try to pass in everything it needs as arguments rather than having it talk up to this.props from inside itself.  This isn't always possible/practical, but since, in this case, you'll be calling this method in `render()`, and you're already destructuring `this.props` to grab `visible`, you might as well grab `navLinks` too, and pass it into your new `getNavLinks` method.

src/components/site/tableOfContents/table-of-contents.module.scss
```diff
@@ -0,0 +1,36 @@
+:global {
+  @include react-md-toolbars;
```
blnkt: probs not necessary if not sticking a toolbar inside the drawer?

src/components/site/tableOfContents/table-of-contents.module.scss
```diff
@@ -0,0 +1,36 @@
...
+  @include react-md-drawers;
+  @include react-md-lists;
+  @include react-md-buttons;
+  @include react-md-icons;
```
blnkt: probs not necessary if using our already styled button component (see comment on line 4)

src/layouts/index.jsx
```diff
@@ -4,33 +4,53 @@ import GlobalStore from '../state/GlobalStore';
...
     store.addReducers();
+
+    this.state = {
+      openSidebar: false,
+    };
   }
 
+  toggleSidebar = () => {
+    const { openSidebar } = this.state;
+    this.setState({ openSidebar: !openSidebar });
```
blnkt: Please replace with a future-proofed setState that incorporates whatever parts of state object we're not touching:
```
this.setState(prevState => ({
   ...prevState,
   openSidebar: !openSidebar,
}));
```
that way if this state object gets more complicated in the future this method won't bulldoze what's there.

src/components/site/tableOfContents/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+
+  componentDidMount() {
+    const { navLinks } = this.props;
+    const links = routes.concat(navLinks);
+    this.navLinks = links.map(link => {
+      if (link.divider || link.subheader) return link;
+
+      if (link.to) {
+        link.component = Link;
+      }
```
blnkt: If you loop over the data that came in through your navLinks prop first, and then concat it to your fixture routes (Home and StyleGuide) then I don't think you need these checks.  Well, I guess you'll also have to add `component: Link` to your fixture routes, but might as well.

src/components/site/tableOfContents/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+
+      if (link.slug) {
+        link.component = Link;
+        link.label = link.title;
+        link.to = `${TO_PREFIX}${link.slug}`;
+        link.primaryText = link.title;
+        link.active = false;
+      }
+      return link;
```
blnkt: this is totally fine, just wanted to point out another spread operator possibility:
```
return ({
   ...link,
   component: Link,
   label: link.title,
   to: `${TO_PREFIX}${link.slug}`,
   primaryText: link.title,
   active: new condition and stuff,
})
```
I haven't incorporated your `link.slug` check, but I'm not totally sure if that's necessary here.  I mean if it's not present it's the map function isn't like gonna throw out that item so you'd still end up passing stuff to the `navItems` prop that potentially won't render correctly so....yea, I don't know, I like the safe-guarding thinking, but I'm not sure the check is necessary at this juncture or doing the job you want it to.

src/components/site/header/index.jsx
```diff
@@ -1,40 +1,59 @@
 import { Link } from 'gatsby';
 import PropTypes from 'prop-types';
 import React from 'react';
-import Toolbar from 'react-md/lib/Toolbars/Toolbar';
+import { Button, Toolbar } from 'react-md';
```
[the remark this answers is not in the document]
georgegomez144: I was originally going to do that, but decided to use the native import because of the use case on the page.

On a side note, trying to import our updated version of the button, but i'm getting an error: a render issue in the Header component.

src/components/site/tableOfContents/index.jsx
georgegomez144: moved into class

src/components/site/tableOfContents/index.jsx
georgegomez144: woah! That's right! I completely forgot about this beautify little gem provided from those spread operators....really need to start utilizing those more.

src/components/site/tableOfContents/index.jsx
georgegomez144: I can almost visualize what you're referring to. I'll have to wrap my mind around the Reacty way of doing the stuffs. I want to and I am going to use this method. Just need to clearly visualize it first.

src/components/site/tableOfContents/table-of-contents.module.scss
georgegomez144: ah, yes. just need a little clean up

src/layouts/index.jsx
georgegomez144: I like this. Need to adapt my development style to the React way of life

src/components/site/tableOfContents/index.jsx
georgegomez144: updated: added the `component: Link` directly in the `this.routes` and spread that over `...this.navLinks.map` to merge them.

src/components/site/tableOfContents/index.jsx
georgegomez144: I get what you're saying here. So, here's what I did:

`componentDidMount() {
    const { navLinks } = this.props;
    this.navLinks = [
      ...this.routes,
      ...navLinks.map(link => {
        if (link.divider || link.subheader) return link;

        if (link.slug) {
          link.component = Link;
          link.label = link.title;
          link.to = `${this.TO_PREFIX}${link.slug}`;
          link.primaryText = link.title;
          link.active = false;
        }
        return {
          ...link,
          component: Link,
          label: link.title,
          to: `${this.TO_PREFIX}${link.slug}`,
          primaryText: link.title,
          active: false,
        };
      }),
    ];
  }`

src/components/site/tableOfContents/index.jsx
georgegomez144: Here is how I resolved this:

```
  componentDidMount() {
    this.updateNavLinks();
  }

  componentDidUpdate() {
    this.updateNavLinks();
  }

  updateNavLinks = () => {
    const { navLinks } = this.props;
    const massagedNavLinks = this.getNavLinks(navLinks);
    this.navLinks = [...this.routes, ...massagedNavLinks];
  };

  setActivePage = linkId => {
    return linkId === this.global.pageId;
  };

  getNavLinks = navLinks => {
    return navLinks.map(link => {
      if (link.divider || link.subheader) return link;
      return {
        ...link,
        component: Link,
        label: link.title,
        to: `${this.TO_PREFIX}${link.slug}`,
        primaryText: link.title,
        active: this.setActivePage(link.id),
      };
    });
  };
```

src/components/site/tableOfContents/index.jsx
georgegomez144: Final output per the previous solution:

```
  render() {
    const { TEMPORARY } = Drawer.DrawerTypes;
    const { visible } = this.props;

    return (
      <Drawer
        type={TEMPORARY}
        className={styles.tableOfContents}
        visible={visible}
        onVisibilityChange={this.handleVisibility}
        navItems={this.navLinks}
        overlay
      />
    );
  }
```

PR #13, opened 2019-11-25
EPO-2361 & EPO-2322

georgegomez144:
This pull request also contains code/work for EPO-2322.

### DIFF
--- a/src/components/site/tableOfContents/index.jsx
+++ b/src/components/site/tableOfContents/index.jsx
@@ -6,6 +6,7 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import { Drawer } from 'react-md';
 import styles from './table-of-contents.module.scss';
+import Check from '../icons/Check';
 
 @reactn
 class TableOfContents extends React.PureComponent {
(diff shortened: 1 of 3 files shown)


src/components/site/tableOfContents/index.jsx
```diff
@@ -40,11 +46,23 @@ class TableOfContents extends React.PureComponent {
         label: link.title,
         to: `${this.TO_PREFIX}${link.slug}`,
         primaryText: link.title,
+        leftIcon: <Check />,
         active: this.setActivePage(link.id),
+        className: classnames(
+          'toc-link',
+          `link--page-id--${link.id}`,
+          this.setActivePage(link.id) ? 'link-active' : '',
```
blnkt: no need to evaluate `setActivePage` twice.  Or rather, if you need a value twice, I'm of the opinion that it's worth setting it to a variable and then just using that variable twice

src/components/site/tableOfContents/index.jsx
```diff
@@ -40,11 +46,23 @@ class TableOfContents extends React.PureComponent {
...
   };
 
+  checkQAProgress = pageId => {
+    // add some fancy logic to reflect page answer state.
+    return +pageId === 5;
+  };
+
   getNavLinks = navLinks => [...this.routes, ...this.massageNavLinks(navLinks)];
 
   setActivePage = linkId => linkId === this.global.pageId;
```
blnkt: consider renaming this method to something like `isActivePage` or `checkActivePage` and reserving the `set` or `get` prefixes for methods that alter state or return something.  Your method is an evaluation, and it's just returning a boolean, a check, whereas, I think one would expect setActivePage to either change what page is active in state, or return "the active page."

src/components/site/tableOfContents/index.jsx
```diff
@@ -40,11 +46,23 @@ class TableOfContents extends React.PureComponent {
...
+          `link--page-id--${link.id}`,
+          this.setActivePage(link.id) ? 'link-active' : '',
+          this.checkQAProgress(link.id) ? 'qa-progress--complete' : ''
+        ),
       };
     });
   };
 
+  checkQAProgress = pageId => {
+    // add some fancy logic to reflect page answer state.
```
blnkt: yes.  The fanciest.

PR #25, opened 2020-01-08
GalaxySelectorComponent

georgegomez144:


src/containers/GalaxySelectorContainer.jsx
```diff
@@ -104,15 +104,26 @@ class GalaxySelectorContainer extends React.PureComponent {
...
-        primaryText: item.name,
+        leftAvatar: <Star style={{ fill: color }} />,
+        primaryText: name,
         className: classnames(
           'galaxy-item',
           'link-item',
-          isActive ? 'link-active' : ''
+          isActive ? 'link-active' : '',
+          isComplete ? 'link-is-complete' : 'link-is-not-complete',
+          isDisabled ? 'link-is-disabled' : ''
```
blnkt: If we're going to use module styles we should go ahead and use them as intended.  As far as I know, you only need to use the global flag to wrap the styles you need to expose in an un-hashed classname (i.e. those classnames that we're trying to override from the material ui library).  For your defined styles (link-active, link-is-complete, link-is-disabled, galaxy-item, etc.) you should alias the import at the top of the file as something (i.e. import styles from blah/blah/blah) and then reference your classnames using camelCase (i.e. styles.linkIsDisabled, styles.linkActive, styles.galaxyItem, etc.) and I believe everything will be copacetic

src/data/pages/interpreting-hubble-plot-1.json
```diff
@@ -19,6 +19,7 @@
     "sources": [
```
blnkt: We're going to need this same data (or at least kind of data) across a bunch of different pages, and in fact, probably across the expanding universe investigation and the supernova investigation, so I think I liked your idea about pulling in this data dynamically through axios.  I'm going to do so with the hubble plot so if you'd like, go ahead and make a separate ticket for refactoring your work to use axios, and you can take a look at what I have when I've pushed it up, and keep on keeping on with what you have so far.

src/components/charts/galaxySelector/galaxySelectorUtilities.js
```diff
@@ -0,0 +1,20 @@
...
+  return originalName.split('_sci.')[0];
+};
+
+export const getNameFromImage = image => image.id;
+
+export const getActiveIndex = (images, activeId) => {
+  return findIndex(images, image => {
+    return getNameFromImage(image) === activeId;
+  });
+};
```
blnkt: please remove any unused functions from the files you copied in

src/data/pages/interpreting-hubble-plot-1.json
```diff
@@ -19,6 +19,7 @@
...
       {
         "name": "ZTF19abqmpsr",
+        "color": "#1ab579",
```
blnkt: Also I think these colors should probably be applied to galaxies within the app and not within the data.  Probably in the container for all of our bits and bobs we can just add this field to the galaxies we're showing. and then pass that updated dataset down to any component that needs it.

src/containers/galaxy-selector-container.scss
```diff
@@ -3,36 +3,43 @@
     padding: 9px;
```
blnkt: Please run the style linter on your new style files and fix and warnings or errors that pop up

src/data/pages/interpreting-hubble-plot-1.json
georgegomez144: Do I not assign the ticket to this sprint. I'm 100% positive that the refactor you're speaking about is going to take a few days.

blnkt: I don't think it's as big a lift as you fear.  As I said, I'm already using axios for the data powering my hubble plot and seems fine.  After you finish the next chunk of work that you would like to add to this pull request we can sit down together and walk through it.  At that point it'll probably be time to merge this branch into master as well.  I don't want our work to stray too far apart since the next next step is going to be integrating your selector and my hubble plot.

src/containers/GalaxySelectorContainer.jsx
georgegomez144: The way the scss file is set up with these class name more or less prohibits me from doing this. Primarily because of the two different nav bars (mini and regular) are styled differently from each other. If you wish, you can review the scss file to see what I'm speaking about.

src/components/charts/galaxySelector/galaxySelectorUtilities.js
georgegomez144: I believe the only function that I'm not using from this file any longer is the getFluid function. I have removed it and will be pushing that change up with my most recent commit.

src/data/pages/interpreting-hubble-plot-1.json
georgegomez144: Wouldn't this restrict the CMS user to not be able to have the flexibility of changing the color representing the galaxies? I added the color prop to the json structure to be able to have it seemlessly carry over into the plotter since the plotted galaxy points will match the color in the sidebar and selection circles. If you truly feel that this functionality is too verbose then I'll be happy to add these colors as static colors into the containing component as suggested.

src/containers/galaxy-selector-container.scss
georgegomez144: Felt weird to duplicate lines in css in order to abide by this "no more that 5 nesting depth" rule, but it is fixed.

src/data/pages/interpreting-hubble-plot-1.json
georgegomez144: Reviewing what you've done in your axios example, I dont feel that another ticket would be necessary. I've already adjust the code to do this.

PR #45, opened 2020-01-27
EPO-2557 Add the photo from the Investigation Doc to the Discuss + Report page

georgegomez144:


src/containers/PageContainer.jsx
```diff
@@ -63,7 +63,7 @@ class PageContainer extends React.PureComponent {
     const options = widget ? widget.options : null;
     const Tag = this.layouts[layout || 'default'];
     const WidgetTag = widget ? this.widgetTags[widget.type] : null;
-    const MediaTag = WidgetTag || image;
+    const MediaTag = WidgetTag; /*  || image */
```
blnkt: Instead of setting MediaTag = WidgetTag, let's go all in on WidgetTag.  That is to say, let's pass the layout components a prop called WidgetTag instead of MediaTag and get rid of this line altogether.

src/components/page/page.module.scss
```diff
@@ -53,4 +54,14 @@
   .grid-media {
     grid-area: grid-media;
   }
+
+  .grid-image {
+    grid-area: grid-image;
+
+    img {
+      width: 100%;
+      height: auto;
```
blnkt: don't need it.  Height is implicitly set to auto if width is explicitly set (as in this case).

src/components/page/page.module.scss
```diff
@@ -42,6 +42,7 @@
     'grid-table-top' auto
     'grid-placeholder' auto
     'grid-media' auto
```
blnkt: probably to go along with our ditching `MediaTag` in the components, let's change these `grid-media`s to `grid-widget`s

src/components/page/page.module.scss
```diff
@@ -53,4 +54,14 @@
   .grid-media {
     grid-area: grid-media;
   }
```
blnkt: probably to go along with our ditching `MediaTag` in the components, let's change these `grid-media`s to `grid-widget`s

src/containers/PageContainer.jsx
georgegomez144: Change made. As I was making these updates, I came to the same conclusion, but left it just in case there was some special case reason that you had created it like this.

src/components/page/page.module.scss
georgegomez144: Check. quick fix.

src/components/page/page.module.scss
georgegomez144: done and done

src/components/page/page.module.scss
georgegomez144: done and done

PR #58, opened 2020-02-07
Bug fixes for multiple tickets

georgegomez144:
Encompasses:

- https://jira.lsstcorp.org/browse/EPO-2613
- https://jira.lsstcorp.org/browse/EPO-2610
- https://jira.lsstcorp.org/browse/EPO-2611
- https://jira.lsstcorp.org/browse/EPO-2615
- https://jira.lsstcorp.org/browse/EPO-2618
- https://jira.lsstcorp.org/browse/EPO-2630

### DIFF
--- a/src/components/charts/galaxySelector/blinker/BlinkerControls.jsx
+++ b/src/components/charts/galaxySelector/blinker/BlinkerControls.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Button from '../../../site/button';
+import ButtonIcon from '../../../site/button/ButtonIcon';
 import Rewind from '../../../site/icons/Rewind';
 import FastForward from '../../../site/icons/FastForward';
 import Pause from '../../../site/icons/Pause';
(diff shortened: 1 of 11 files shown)


src/components/charts/galaxySelector/blinker/BlinkerControls.jsx
```diff
@@ -20,19 +21,19 @@ const BlinkerControls = ({
...
         icon
-        iconEl={<StartStopTag />}
+        iconEl={<ButtonIcon srText="Start/Stop" Icon={StartStopTag} />}
         onClick={handleStartStop}
       />
       <Button
         className={control}
         icon
-        iconEl={<FastForward />}
+        iconEl={<ButtonIcon srText="Fast Forward" Icon={FastForward} />}
```
blnkt: This seems to be working well.  THe other spot that could use this treatment is the mini NavDrawer menu items.  Those all just read out as "Toggle button", as well as the button that opens and closes the full sized NavDrawer.  The latter seems to have a span inside it that should read out in a screen reader correctly, but doesn't for some reason. The case of the former is more critical than the latter as the fullsized NavDrawer is not in fact accessible at all, but that's my bad, an oversight I'll have to deal with at a later date.  But for now, if possible, I think we'd like to see those icon only NavDrawer items get screen reader only text if at all possible within the confines of our react-md world.

src/components/charts/shared/navDrawer/nav-drawer.module.scss
```diff
@@ -1,6 +1,7 @@
 .nav-drawer-container {
   position: relative;
   overflow: hidden;
+  z-index: 1;
```
blnkt: z-index should come before overflow (see order/properties-order linter warning in terminal)

src/components/charts/shared/observationsTables/ObservationsTable.jsx
```diff
@@ -59,6 +59,7 @@ class ObservationsTable extends React.PureComponent {
       <div className={this.getPositionClassname(position)}>
         {title && <h1 className={styles.tableTitle}>{title}</h1>}
         <Table
+          tabindex="0"
```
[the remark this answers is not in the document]
georgegomez144: All set

src/components/charts/shared/navDrawer/nav-drawer.module.scss
georgegomez144: ran `yarn styles`

PR #60, opened 2020-02-11
EPO-2626 Positionable Widgets (Widgets should support optional "layout" field)

georgegomez144:


src/components/page/index.jsx
```diff
@@ -16,6 +16,9 @@ class Page extends React.PureComponent {
       tables,
       image,
       WidgetTag,
+      widget: {
+        layout: { row: widgetRow },
+      },
```
blnkt: This is what's causing the build to fail: not all pages have widgets (and not all widgets have layout fields on them for that matter), so sometimes widget is null, so sometimes this line is attempting to destructure null which is a non-starter.  Probably the best way to address this would be to break this line into multiple steps:
```
const { layout } = widget || {};
const { row: widgetRow } = layout || {}; 
```
And then deal with the fact that sometimes widgetRow will be null.

src/components/page/index.jsx
```diff
@@ -35,16 +43,16 @@ class Page extends React.PureComponent {
...
         {image && (
           <div className={styles.gridImage}>
             <img src={image.mediaPath} alt={image.altText} />
           </div>
         )}
+        {widgetRow === 'bottom' && WidgetTag && (
+          <div className={styles[`gridWidget${capitalize(widgetRow)}`]}>
+            <WidgetTag {...this.props} />
+          </div>
+        )}
```
blnkt: I think this is probably where we want a widget to end up by default so I think you could change line 51 to something like: `(widgetRow === 'bottom' || !widgetRow) && WidgetTag && `

I think...

src/data/pages/lsst-hubble-plot.json
```diff
@@ -17,6 +17,10 @@
   "widget": {
     "type": "HubblePlot",
     "source": "/data/galaxies/multiple_galaxies.json",
+    "layout": {
+      "col": "left",
+      "row": "bottom"
+    },
```
blnkt: I don't see any reason why the positioning attributes shouldn't be consistent across our main page components: images, tables, and widgets.  As I understand it you are using a competing may to define position and to build the classname that kicks off your css grid stuff.  As you've done this in two different ways now, you probably have a sense of which mode is working better for ya: whichever mode that is, please update your work so it's consistent across all currently positionable components.

georgegomez144: I really do like this current solution better that then CSS-Grid solution. That IE issue is going to always dampen development until they completely destroy that app and no one, and I mean, no one uses IE any longer. I'm just going to have to update the tables work that I did with CSS-Grid and implement this solution.

PR #66, opened 2020-02-17
EPO-2562 Add "multi-select" question type

georgegomez144:


src/components/qas/questions/qaMultiSelect/qaMultiSelect.module.scss
```diff
@@ -0,0 +1,69 @@
+:global {
+  @include react-md-menus-media;
+  @include react-md-selection-controls;
+
```
blnkt: I think it behooves us to use this global trick as exclusively as we can.  And in my testing wrapping only these includes in global still yielded the styling we'd expect.

src/components/qas/questions/qaMultiSelect/qaMultiSelect.module.scss
```diff
@@ -0,0 +1,69 @@
...
+  :global .md-list--menu {
+    position: absolute;
+    z-index: 1;
+  }
+
+  .selected-options {
+    height: 20px;
+    font-weight: 200;
+    color: #222222;
```
blnkt: I prefer that as much as possible colors are sass variables so they're easier to keep track of and we don't let the project start using dozens of almost the same colors, so let's use $black instead of #222222 in both of these cases here

src/components/qas/questions/qaMultiSelect/qaMultiSelect.module.scss
```diff
@@ -0,0 +1,69 @@
...
+  position: relative;
+
+  :global .md-list--menu {
+    position: absolute;
+    z-index: 1;
+  }
+
+  .selected-options {
+    height: 20px;
+    font-weight: 200;
```
blnkt: This font doesn't actually have a 200 weight, but I think $regular is what other text like this is using

src/components/qas/questions/qaMultiSelect/qaMultiSelect.module.scss
```diff
@@ -0,0 +1,69 @@
...
+    .label-post {
+      @include copySecondary;
+      line-height: $baseLineHeight;
+    }
+
+    select {
+      @include copyPrimary;
+    }
+
+    &:not(.answerable) {
```
blnkt: this not answerable styling is working great, but please add a rule to make the text (like what would be the answer, but what starts out as the "Select" placeholder text) $neutral20 when .qa-select:not(.answerable).  This is easier to test and see if you temporarily make this multiselect question not the first question on the summary page

src/components/qas/questions/qaMultiSelect/index.jsx
```diff
@@ -0,0 +1,197 @@
...
+      label,
+      labelPre,
+      labelPost,
+      options,
+      placeholder,
+      questionType,
+    } = question;
+
+    const active = ids ? checkIds(ids, activeId) : activeId === id;
+    const { hasFocus, answerable, selectedOptions } = this.state;
```
blnkt: The toggle that opens this menu should be disabled (either literally or as close an approximation as we can muster) is answerable is false

src/components/qas/questions/qaMultiSelect/index.jsx
```diff
@@ -0,0 +1,197 @@
...
+  getMultiSelectOptions = options => {
+    const { selectedOptions } = this.state;
+    return options.map((option, i) => {
+      return (
+        <SelectionControl
+          id={`qa-multi-select--${option.value}`}
+          checkedCheckboxIcon={<CheckBox />}
+          uncheckedCheckboxIcon={<CheckBoxOutlineBlank />}
+          // eslint-disable-next-line react/no-array-index-key
+          key={`${option.label}-${i}--option`}
```
blnkt: I think just take out the index from this key value and then you don't need to disable this rule.  It would I think never ever be the case that the label to two select options will be the same because then why would you have both options.

georgegomez144: Got it. Yeah, my editor looks like a crime scene when I take out the key, but as long as the app doesn't crash, haha

src/components/qas/questions/qaMultiSelect/qaMultiSelect.module.scss
georgegomez144: Not sure if this is an action item or not.

src/components/qas/questions/qaMultiSelect/index.jsx
georgegomez144: I'm not sure I follow you on this. I am a little confused as to what you would like me to do with this.

src/components/qas/questions/qaMultiSelect/qaMultiSelect.module.scss
georgegomez144: added a `pointer-events: none;` as the fix for this

src/components/qas/questions/qaMultiSelect/index.jsx
georgegomez144: for now, handling this in the CSS

src/components/qas/questions/qaMultiSelect/index.jsx
```diff
@@ -0,0 +1,197 @@
...
+        }),
+        () => {
+          if (focusCallback) {
+            const { hasFocus } = this.state;
+            focusCallback(hasFocus);
+          }
+          const { selectedOptions } = this.state;
+          handleAnswerSelect(question.id, selectedOptions, e.type);
+          saveHandler();
+        }
```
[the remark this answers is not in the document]
georgegomez144: all complete. For number 4, I just changed the `e.type` to simple `e` and found no errors.

src/components/qas/questions/qaMultiSelect/qaMultiSelect.module.scss
georgegomez144: 👍

PR #68, opened 2020-02-18
Pages Support Multiple Positionable Images

georgegomez144:
EPO-2648
EPO-2649


blnkt:
Please update pageData to use new "images" field instead of "image" field.  This will allow us to better evaluate if stuff is working where we would expect, which is pivotal, and also not break all of the things when we merge it in...which is also kinda important and stuff.

src/components/page/shared/imageBlock.jsx
```diff
@@ -0,0 +1,19 @@
...
+    )
+  );
+}
+
+ImageBlock.propTypes = {
+  image: PropTypes.object,
+  classname: PropTypes.string,
+};
+
+export default ImageBlock;
```
blnkt: This is a great start for what do in [EPO-2704](https://jira.lsstcorp.org/browse/EPO-2704).  I think we can imagine something like a blocks directory in components directory where the index file will have the logic for deriving the grid classnames and then what kind of Block to render where then we could have these pure presentational components like you have with ImageBlock (so like also a WidgetBlock and a TableBlock or whatev) and then the layout components (TwoCol and SingleCol) won't have to have all these renderStuffs methods, and probably won't even need to know about the different types of Widgets either as we can derive the WidgetTag in the same loop as the grid classname.

PR #74, opened 2020-03-04
EPO-2758 Update basic brightness plot to support plotting color vs Distance plot

georgegomez144:
Work continued from EPO-2757


src/components/charts/brightnessVsDistance/Points.jsx
```diff
@@ -35,6 +35,8 @@ class Points extends React.PureComponent {
             [notActive]: (selectedData || hoveredData) && !selected && !hovered,
             [invisible]: x === null || y === null,
           });
+          const blueColorPercent = Math.floor((color / 2) * 100);
+          const redColorPercent = 100 - blueColorPercent;
```
blnkt: Great trick.  I love the results, and so does Lauren.  I asked her to think about if there are any sciency color cutoffs we should consider, or sciency hues of red and blue.  If we end up wanting to assert even more control over how to translate these values into visual colors I'm sure we can abstract the work I did in the HR Diagram to translate temps into colors (see `charts/scatter-plot/points.jsx`)

src/data/pages/exploring-lsst-data-1.json
```diff
@@ -23,6 +23,7 @@
       },
       "options": {
         "title": "Color Vs Distance",
+        "showFlux": true,
```
blnkt: This is a sensible strategy to get this boolean where you need it, but I really think this is a bit of a slippery slope.  We've done a good job of abstracting the BrightnessVsDistance plot to be a GalacticPropertiesPlot but I think it's still better to rely on the x and y valueAccessors and let those inform point styling.  So you could even still use this `showFlux` flag internal to the components if you want, but I feel like `showFlux` is a little out of place in the pageData.  Instead let's just set the flag to true is the yValueAccessor is 'color'.

src/components/charts/brightnessVsDistance/Points.jsx
georgegomez144: Thank you!!! I was more than excited when I put this together. Very happy with the results.

src/data/pages/exploring-lsst-data-1.json
georgegomez144: Okay, removed showFlux from the data and all places it 'touched' on it's way into the Points.jsx

All should be good. I've restarted and cleaned the gatsby to check for error with none found.

PR #80, opened 2020-03-10
EPO-2751 Update GalaxySelector to accept explicit image path instead of programmatically determined image stack based on alerts

georgegomez144:


src/components/charts/galaxySelector/index.jsx
```diff
@@ -337,13 +338,29 @@ class GalaxySelector extends React.PureComponent {
...
-          />
+          {image ? (
+            <img
+              style={{
+                width: '100%',
+                position: 'absolute',
+                top: 0,
+                left: 0,
+                zIndex: '-1',
+              }}
```
blnkt: Please encapsulate these styles in a new class in the `galaxy-selector.module.scss` file.  Also `z-index: -1` is not necessary in this case as the `img` element appears after the intractable portions of the widget (which is also absolutely positioned) and so is already correctly placed in the DOM flow (i.e. the `img` is already "under" the `svg` -- which is the clickable part of this widget).

georgegomez144: done and done :)

PR #84, opened 2020-03-13
[EPO-2752] GalaxiesSelector supports selections

blnkt:
Includes minor updates to GalaxySElector & GalacticProperties styles;
Includes update to QAAccordion padding to match other card-style QAs;
Includes updated hsc data;
Includes qa content update to count selected galaxies;
Includes some duplicated changes from previous commit (accidental);


src/components/charts/galacticProperties/Points.jsx
```diff
@@ -37,7 +37,7 @@ class Points extends React.PureComponent {
           });
           const blueColorPercent = Math.floor((color / 2) * 100);
           const redColorPercent = 100 - blueColorPercent;
-          const colorShift = `rgb(${redColorPercent}%, 0%, ${blueColorPercent}%)`;
+          const colorShift = `rgba(${redColorPercent}%, 0%, ${blueColorPercent}%, 0.85)`;
```
georgegomez144: Love the use of the transparency here!!!!

src/containers/GalaxiesSelectorContainer.jsx
```diff
@@ -43,19 +47,47 @@ class GalaxiesSelectorContainer extends React.PureComponent {
...
+    );
+
     return (
       <>
         <h2 className="space-bottom">Galaxies Selector</h2>
         <div className="galaxies-selector-images--container">
           <GalaxySelector
             className="galaxies-selector"
+            {...{ selectedData }}
             data={data}
```
georgegomez144: Why not include `data={data}` in the spead above??? something like `{...{ selectedData, data }}`??

src/containers/WithQAing.jsx
```diff
@@ -105,6 +106,13 @@ export const WithQAing = ComposedComponent => {
       return `${numOfGalaxies} galaxies & ${numOfSupernovae} supernovae`;
     }
 
+    getGalaxiesContent(data) {
+      const numGalaxies = Object.keys(data).length;
+      return `${numGalaxies} ${
+        numGalaxies > 1 ? 'galaxies' : 'galaxy'
```
georgegomez144: Maybe also do `numGalaxies > 1 || numGalaxies === 0 ? 'galaxies' : 'galaxy'`
Not sure if we'll ever have 0 galaxies, but at least if we did the text would show correctly?

blnkt: Good catch, but in this case content is not rendered unless there is an answer present (a selection), and you are not able to "select" 0 things so we can confidently assume numGalaxies >= 1

src/containers/GalaxiesSelectorContainer.jsx
blnkt: true enough: will make this change and amend it to [EPO-2790](https://github.com/lsst-epo/investigation-static/pull/85)

src/containers/WithQAing.jsx
georgegomez144: Awesome! That's what I figured, but just wanted to be certain.

PR #85, opened 2020-03-13
[EPO-2790] GalaxiesSelector selections populate GalacticProperties plot

blnkt:
POC uses brightness vs distance plot explicitly;
This pull request uses work on another branch: [EPO-2752](https://github.com/lsst-epo/investigation-static/pull/84)


src/components/charts/galaxySelector/Point.jsx
```diff
@@ -9,7 +9,7 @@ class Point extends React.PureComponent {
       {
         supernova: 20,
         galaxy: 40,
-      }[id] || 5
+      }[id] || 8
```
georgegomez144: Help me understand this return of `8` if the index requested does not exist...

blnkt: surely.  So, we've got this object literal `{ supernova: 20, galaxy: 40 }` and in the case of the GalaxySelector, as implemented in the Expanding Universe Investigation, id is either galaxy or supernova, since we like made that data ourselves, and we wanted a big circle around the galaxy and a small one around the supernova.  In the case of the HSC data, the ids are numbers, so because those keys don't exist in this object it'll return undefined which is falsey so it'll return the other side of the or.  It's very similar to the strategy we used for the QAs, blocks, and page layouts.

blnkt:
@georgegomez144 right you are.  I split the plotting and the question answering onto separate pages but forgot to add "preSelected" option to the plot and GalaxiesSelector on `exploring-the-deep-1` (they're done plotting so should just be looking and clicking should do nothing except highlight -- when I finish EPO-2794).  I'll fix this right now and amend to last commit on this pull request

src/components/charts/galaxySelector/Point.jsx
georgegomez144: I'm guessing the value of `8` holds significance to the response?

georgegomez144:
Ahh, very cool. That make much more sense!

src/components/charts/galaxySelector/Point.jsx
blnkt: 20, 40, and 8 are the radii of points

PR #87, opened 2020-03-17
[EPO-2795] GalaxiesSelector and GalacticProperties components support randomized data sources

blnkt:
Includes support for user generated data alongside random data source or sources;
Includes updated pageData implementing random/multi data sources;
Random data source saved as "hidden" answer (answer wo question);
Includes updated (compressed) hsc images;


src/containers/GalaxiesSelectorContainer.jsx
```diff
@@ -33,7 +57,7 @@ class GalaxiesSelectorContainer extends React.PureComponent {
       this.setState(prevState => ({
         ...prevState,
         data: objects,
-        imagePath: `/images/galaxies/hsc/${name}.png`,
+        imagePath: `/images/galaxies/hsc/${name}.jpg`,
```
georgegomez144: I'm understanding that from this, there will always be a compressed jpg file format and not a png?

src/state/GlobalStore.js
```diff
@@ -124,7 +124,18 @@ class GlobalStore {
...
-        const { questions, answers: prevAnswers } = prevIPS[pageId];
+        const prevPageIPS = prevIPS[pageId];
+
+        if (!prevIPS[pageId]) {
+          return global;
+        }
+
+        const { questions, answers: prevAnswers } = prevPageIPS;
+
+        if (!questions[qId]) {
```
georgegomez144: On the page `/observable-universe/exploring-the-deep/`, I'm getting an error in the console (page not loading futher) that says, "Cannot read property 'randomGalaxies' of null" that points to this line.

src/containers/GalaxiesSelectorContainer.jsx
blnkt: sure.  In this case the pngs were just way too big and were slowing page loads down so I elected to update the files used.

src/state/GlobalStore.js
blnkt: I'm not able to reproduce the issues you're describing.  Are you on the latest commit? Can you provide instructions to reproduce `"Cannot read property 'randomGalaxies' of null"` and `/observable-universe/exploring-the-deep-1/ is 'stuck' loading the page. Only the spinner is loading on the selector.`?

src/containers/GalaxiesSelectorContainer.jsx
georgegomez144: 👏 Good decision, then!!!!

src/state/GlobalStore.js
georgegomez144: No specific instructions other than go to the page. I did nothing other than clear my local storage.

PR #90, opened 2020-03-18
[B] Fix globalState race condition in randomData

blnkt:
Previously, when random data option was true GalaxiesSelector fetched random data and then immediately created an answer in order to persist the data across pages.  That was causing a race condition between the fetch and GlobalState updating when visiting the page for the first
with nothing previously saved in localStorage.  This update punts the answer creation to DidUpdate avoiding since the answer creation is not vital until after you click through to a different page.  This avoids the race altogether

### DIFF
--- a/src/containers/GalaxiesSelectorContainer.jsx
+++ b/src/containers/GalaxiesSelectorContainer.jsx
@@ -15,22 +15,23 @@ class GalaxiesSelectorContainer extends React.PureComponent {
     this.state = {
       data: null,
       name: null,
+      sourcePath: null,
       imagePath: null,
       domain: [],
       activeGalaxy: null,
     };
+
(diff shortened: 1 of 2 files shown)


blnkt:
@georgegomez144 I did not ask you to review this pull request because it is in fact addressing an issue you brought up in a [comment on a previous pull request](https://github.com/lsst-epo/investigation-static/pull/87#pullrequestreview-376192882).  At that time we looked at it together and agreed the inconsistent issue we were seeing was related to our browsers and not our code.  As such we approved the pull request and merged it.  Since then, I found a way to consistently reproduce the bug, looked more carefully at the code and discovered an easy workaround, and so addressed it here.  I tested it thoroughly in my browser, I'm satisfied with the results, and so didn't feel like this change warranted your review.

PR #91, opened 2020-03-19
[EPO-2754] Add GalacticPropertiesCombo component.

georgegomez144:
Includes a new container component to house nav and plot
Includes a way to switch between brightness and color views using
  the same data source.


src/containers/GalacticPropertiesComboContainer.jsx
```diff
@@ -0,0 +1,203 @@
...
+        : this.getColorOptions();
+    this.setState(prevState => ({
+      ...prevState,
+      navItems: this.generateNavItems(properties, property),
+      options: loadOptions,
+      updateDomain: true,
+    }));
+  };
+
+  generateNavItems(properties, activeScramble) {
```
blnkt: clean up this `activeScramble` variable name.  That was specific to the galaxyScrambler.

src/containers/GalacticPropertiesComboContainer.jsx
```diff
@@ -0,0 +1,203 @@
...
+      return {
+        leftAvatar: (
+          <span>
+            <ScatterPlot style={{ fill: color }} />
+            <span className="screen-reader-only">{name}</span>
+          </span>
+        ),
+        primaryText: name,
+        className: classnames(propertyItem, {
+          [linkActive]: activeScramble.name === property.name,
```
blnkt: clean up this `activeScramble` variable name.  That was specific to the galaxyScrambler.

src/containers/GalacticPropertiesComboContainer.jsx
```diff
@@ -0,0 +1,203 @@
...
+                  data,
+                  options,
+                  xAxisLabel,
+                  yAxisLabel,
+                  xValueAccessor,
+                  yValueAccessor,
+                  tooltipAccessors,
+                  tooltipUnits,
+                  tooltipLabels,
+                  updateDomain,
```
blnkt: And so if you made the changes described above in the GalacticProperties component you wouldn't pass this updateDomain bool through because GalacticProperties would just be responding to other props changing.

src/containers/GalacticPropertiesComboContainer.jsx
```diff
@@ -0,0 +1,203 @@
...
+                  xDomain,
+                  yDomain,
+                }}
+                domainUpdatedCallback={this.handleDomainUpdatedResponse}
```
blnkt: And also you wouldn't need this method in the parent or passed through to the child

src/components/charts/galacticProperties/index.jsx
```diff
@@ -63,12 +63,43 @@ class GalacticProperties extends React.Component {
...
+
+    this.setState(
+      prevState => ({
+        ...prevState,
+        xScale: d3ScaleLinear()
+          .domain(domain ? domain[0] : xDomain)
+          .range([padding, width]),
+        yScale: d3ScaleLinear()
+          .domain(domain ? domain[1] : yDomain)
+          .range([height - padding, 0]),
```
blnkt: DRY. You got the same thing here as above in the constructor.  break it our into one or more methods like getXScale getYScale and then reuse in both spots (see LightCurve component for an example of this)

PR #93, opened 2020-03-23
[C] Use single toggle slider instead of dual toggle.

georgegomez144:
[EPO-2834]
Include an update to only show one toggle.
Range selected is within a one hundredth of the full data at a time.
Currently, returns just the selection.


georgegomez144:
Looks great! I just created that ticket and assigned it to me.

PR #95, opened 2020-03-24
[EPO-2889] GalaxySelector renders activeImage on switch

blnkt:
When activeGalaxy changes, GalaxySelector renders first image of new set and accurately sets the activeAlert from new set;


blnkt:
Associated commit delivered in pull request #96

PR #96, opened 2020-03-24
[EPO-2891] GalaxySelectorHubblePlot advances to next unselected galaxy

blnkt:
When clicking previous/next galaxy buttons below slideout HubblePlot:
- If new activeGalaxy is already selected in GalaxySelector then highlight point in HubblePlot
- If the activeGalaxy is not selected in GalaxySelector then close HubblePlot and update GalaxySelector

### DIFF
--- a/src/containers/GalaxySelectorContainer.jsx
+++ b/src/containers/GalaxySelectorContainer.jsx
@@ -1,4 +1,3 @@
-/* eslint-disable no-console */
 import React from 'react';
 import PropTypes from 'prop-types';
 import isEmpty from 'lodash/isEmpty';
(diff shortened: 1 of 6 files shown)


src/containers/GalaxySelectorContainer.jsx
```diff
@@ -27,7 +27,7 @@ class GalaxySelectorContainer extends React.PureComponent {
     this.state = {
       data: null,
       plottedData: null,
-      openScatterPlot: false,
+      plotIsOpen: false,
```
georgegomez144: A much better name for this variable 👍

PR #99, opened 2020-03-26
[EPO-2836] pull in code and create new custom parent component for the Large Scale Structure (side by side)

georgegomez144:
Structure of the 2d Chart has been massaged and updated to accept the data format currently being consumed by the 3d Chart. Both charts will update according to the slider in the 3d Chart (i.e. the selected data displayed).

Note: may need to break out the slider into the parent container to have more fluid functionality.

### DIFF
--- a/src/components/charts/largeScaleStructurePlot/2DChart.jsx
+++ b/src/components/charts/largeScaleStructurePlot/2DChart.jsx
@@ -1,0 +1,189 @@
+import React from 'react';
+import ReactEcharts from 'echarts-for-react';
+import PropTypes from 'prop-types';
+import './large-scale-structure-plot.module.scss';
+import { formatValue } from '../../../lib/utilities';
+
+class LargeScaleStructure2DPlot extends React.PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.primaryColor = '#374785';
+    this.secondaryColor = '#A8D0E6';
+  }
+
+  formatTooltipPoint = (color, string) => {
+    return (
+      "<span style='display:inline-block;width:10px;height:10px;border-radius:50%;background-color:" +
+      color +
+      ";margin-right:5px;'></span><span>" +
+      string +
+      '</span>'
+    );
+  };
+
+  getTooltipFormat = params => {
+    const data = params.value;
+
+    if (Array.isArray(data)) {
+      const [ra, dec] = data;
+      return this.formatTooltipPoint(
+        this.secondaryColor,
+        `${ra.toFixed(2)}, ${dec.toFixed(2)}`
+      );
+    }
+
+    const { RA, Dec, redshift } = data;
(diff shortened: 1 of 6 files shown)


src/components/charts/largeScaleStructurePlot/2DChart.jsx
```diff
@@ -0,0 +1,302 @@
...
+
+      color: ['#A8D0E6', '#54ff', '#374785'],
+      series: [
+        {
+          name: 'Data Point',
+          symbolSize: 8,
+          type: 'scatter',
+          legendHoverLink: true,
+          large: largeBool,
+          data,
```
blnkt: Axes of graph read Dec and RA, which is, I think, the dimensions we are interested in showing here, however the values don't line up correctly: this graph is definitely plotting RA & Redshift.  ?this is probably related to the fact that the data coming into this component is an array of arrays now whereas the work you copied and pasted this out of was maybe accepting an array of objects before?  I'm not sure.  I think there's a dimensions filed you can use to point the graph to the keys/indexes you want it to plot (I think this is why you see RA and DEC pop up in the dimensions for the dataset option above?  Although this is maybe the syntax you would use to reference keys in an object and so maybe is not doing anything here? ) And in fact I just commented that line out and see no visual or behavioral change so I think what values are getting plotted right now is 100% running off of whatever eCharts grabs by default, and that's not quite lining up with what we have.  I just tried updating the `arrayifyData` method in the container to spit out the values into arrays in a different order, and that solved the issue in the2D plot but messed it up in the 3D plot.  So either here or there something needs to change to better define what values get used where.  And either way I'd like the code to be explicit about what values to use (i.e. if given the option I'd rather see us telling echarts which dimensions of the data to plot rather than changing our data to line up to echarts' default behavior).

src/components/charts/largeScaleStructurePlot/2DChart.jsx
```diff
@@ -0,0 +1,302 @@
...
+class LargeScaleStructure2DPlot extends React.PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      graphData: {},
+      clickedPt: {},
+      selectedPt: [],
+      hover: false,
+      hoverPt: null,
```
blnkt: defined but never used

src/components/charts/largeScaleStructurePlot/2DChart.jsx
```diff
@@ -0,0 +1,302 @@
...
+          name: 'Data Point',
+          icon: 'circle',
+        },
+        {
+          name: legendPt,
+          icon: legendIcon,
+        },
+      ],
+    };
+  };
```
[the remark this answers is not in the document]
georgegomez144: No 100% certain how this bit is working. I tried updating the values to reflect that data instead, but nothing rendered.

PR #100, opened 2020-03-30
[EPO-2846] GalacticProperties chart has legend in presence of multiple datasets

blnkt:
Includes updated GalaxySelector: use radii for sizing objects' selectable areas


src/components/charts/galaxySelector/Points.jsx
```diff
@@ -20,16 +20,18 @@ class Points extends React.PureComponent {
...
           return (
             <Point
               key={`point-${id}`}
-              {...{ id, isSelected, isActive }}
+              {...{ id, isActive, isSelected }}
+              radius={xScale(xVal - modR) - xScale(xVal + modR)}
               classes={pointClasses}
-              x={xScale(d[xValueAccessor])}
+              x={xScale(xVal)}
               y={yScale(d[yValueAccessor])}
```
georgegomez144: why not do the same with the `yValueAccessor` ?

blnkt: no reason I guess. As I rule of thumb I just only assign a value to a variable if it gets used more than once.  Force of habit more than anything I spoze.

PR #107, opened 2020-04-01
[EPO-2791] A good foundation to showcase density.

georgegomez144:


blnkt:
Continuing this work in another ticket so closing this ticket without rebase and merge

PR #108, opened 2020-04-03
[C] Create Size Calculator component

georgegomez144:


src/components/qas/questions/qaCalculators/sizeCalculator/index.jsx
```diff
@@ -0,0 +1,242 @@
...
+  );
+};
+
+const PIcon = () => {
+  return (
+    <div>
+      <strong>p</strong> =
+    </div>
+  );
+};
```
blnkt, line 30: I'm getting some funny
```
h
=
```
 text wrapping at narrow (but still reasonably wide) screen sizes.  I think it's because the `h` and the `=` are both `display: inline`.  I think you can fix by making them both inline-block (i.e. wrapping the = and whitespace in a span and setting it to inline-block, and updating display property of the strong).

src/components/qas/questions/qaCalculators/sizeCalculator/index.jsx
```diff
@@ -0,0 +1,242 @@
...
+              placeholder={placeholder}
+              defaultValue={answered ? albedo : null}
+              onBlur={this.handlePBlur}
+              onFocus={this.handlePFocus}
+              onChange={this.handlePChange}
+              disabled={!(answerable || answered || active)}
+            />
+          </div>
+          <div className="results-list">
+            <CalculatedMeasurement unit="D" value={diameter} />
```
blnkt, line 218: `D` always `= ?` on pageload even when there is an answer present.  Probably just related to the issue described above in the `componentDidMount` method.

src/components/qas/questions/qaCalculators/sizeCalculator/index.jsx
```diff
@@ -0,0 +1,242 @@
...
+  handleChange = (value, index) => {
+    const { question, answerHandler } = this.props;
+    const { id } = question;
+    const { value: oldValue } = this.state;
+    const newVal = {
+      ...oldValue,
+      [index]: value,
+    };
+
+    if (newVal.h && newVal.p && newVal.d) {
```
blnkt, line 107: I am fairly certain this will never be true when you want it to be.  You're checking h, p, and d on the newVal.  I believe this will always yield the newest h and p, but an old d, such that even if the condition is met it will pass through inaccurate values.  

I think probably what you want to do is either `updateCalculatedMeasurements` and then, peak at updated state, and then pass `answerHandler` the updated `value` on the updated state, 
OR
 `calculateDiameter` here, include it in `newVal`, pass it on to `answerHandler`, and update `updateCalculatedMeasurements` to NOT calculate the diameter, but instead just be in charge of updating your state.

PR #116, opened 2020-04-17
[EPO-2926] Add basic histogram component

blnkt:
Implements across Solar System investigation pages;
Does not include support for handling multiple or combo histograms;


blnkt:
work is on #119

PR #117, opened 2020-04-20
[EPO-2927] Support toggling Histogram using NavDrawer

blnkt:
Includes a new component called ChartSwitcher which (untested)
supports any variety or number of widgets inside the existing
NavDrawer component and def supports histograms;


blnkt:
work is on #119

PR #120, opened 2020-04-22
[C] Testing in Jest

georgegomez144:
[EPO-2906] Very basic render test of the distanceCalculator in Jest.

### DIFF
--- a/src/components/qas/questions/qaCalculators/distanceCalculator/__snapshots__/distanceCalculator.test.js.snap
+++ b/src/components/qas/questions/qaCalculators/distanceCalculator/__snapshots__/distanceCalculator.test.js.snap
@@ -1,0 +1,196 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Distance Calculator renders with question #66 1`] = `
+Object {
+  "asFragment": [Function],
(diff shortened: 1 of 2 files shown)


blnkt:
This looks good: you set up a test that makes sense, gave yourself some fixture data to test with, rendered some stuff.  Unfortunately, I think we likely will not be using snapshot tests very extensively in this phase of development.  Totally a straight forward kind of first test to try to write so great, I wanted to jump in here so you knew that you shouldn't focus too much on what jest provides for generating, utilizing, and manipulating snapshots.   Snapshot tests (also known as "smoke test") compare an old "snapshot" of the html react compiles to a new "snapshot" created when the test runs.  Currently we are typically changing our components a lot, and new features are often, by design, updating/changing the UI (the html that is compiled) so these kinds of "smoke tests" tend to produce a lot of false positives and you end up in this habit of updating snapshots every time you run your tests, which more-or-less completely misses the point of having the snapshots in the first place, so might as well pass on it.  If/when you return to more unit testing stuff next sprint I would encourage you to focus more on how to craft tests that are simulating conditions that affect what gets rendered, and tests that simulate human interactions that produce changes to the dom or data flow.

PR #132, opened 2020-05-01
[EPO-3018] Add "end" page to investigation flow

blnkt:
Includes updates to TOC and Progress tracking component

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -201,6 +201,23 @@ exports.createPages = async ({ graphql, actions }) => {
     context: { investigations, env: INVESTIGATION },
   });
 
+  if (isAll) {
+    investigations.forEach(investigation => {
+      const { id } = investigation;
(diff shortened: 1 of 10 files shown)


blnkt:
@georgegomez144 yes, sorry that's to be expected, I should have said something.  The progress tracking is more or less restricted to tracking 1 investigation at a time.  Thus on staging and locally when if you're building all of the investigations you won't see what you expect in the table of contents on the main landing page, or within the investigation if you enter into it using the links on the landing page.  Once you're inside an investigation, if you reload the page you should see what you would expect, and it should progress from there accurately.  As for the end page, since all of the public facing versions of the investigations are 1 build per investigation, I'm only showing a summary message and some links and tools specific to your investigation experience if the user is in one of the public facing environments.  Otherwise you just get the same thing you get on the landing page: links to navigate to any one of the investigations.

PR #135, opened 2020-05-04
[EPO-2982] Add playback controls to Obritviewer

blnkt:


blnkt:
This pull request has been reviewed and approved but the associated commits will be delivered #136 instead of here.

PR #140, opened 2020-05-07
[EPO-3025] Histogram Tooltip pos uses mousePos & chartHeight

blnkt:
Previously was using the position of the bar in the histogram to position tooltip but was causing inconsistencies: using mousePos and top of the chart is way more consistent although less elegant

### DIFF
--- a/src/components/charts/histogram/index.jsx
+++ b/src/components/charts/histogram/index.jsx
@@ -2,7 +2,11 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import isEmpty from 'lodash/isEmpty';
 import classnames from 'classnames';
-import { select as d3Select, event as d3Event } from 'd3-selection';
+import {
+  select as d3Select,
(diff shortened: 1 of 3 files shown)


src/components/charts/histogram/index.jsx
```diff
@@ -230,13 +235,24 @@ class Histogram extends React.PureComponent {
     }));
   }
 
-  toggleSelection(d) {
+  getTooltipPos(data, pos) {
+    const { padding, offsetTop, offsetRight, height, width } = this.props;
+    const { xScale, yScale } = this.state;
+    console.log('x', xScale(data.x0), [padding, width - offsetRight]);
+    console.log('y', yScale(data.length), [height - padding, offsetTop]);
```
georgegomez144, line 242: console.log(s) still in code here

blnkt:
yea I see that lil underneath moment too.  I can't remember if it's always been like that...

PR #141, opened 2020-05-08
[EPO-3033] Create QACompoundTextInput

georgegomez144:
Still struggling with active/inactive states.


src/components/qas/questions/qaCompoundTextInput/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+
+  render() {
+    const { questions, activeId, answers, handleAnswerSelect } = this.props;
+    const { hasFocus } = this.state;
+    const cardClasses = classnames('qa-card', { active: hasFocus });
+    const containerClasses = classnames('qa-wrapper');
+
+    return (
+      <Card className={cardClasses}>
+        <div className={containerClasses}>
```
blnkt, line 32: don't really need this containerClass variable or classnames if you're only giving it a static classname

PR #148, opened 2020-05-13
[EPO-3152] TimeDomainViewer objects are selectable

blnkt:
To test you would need to create a accordion style question with an neo accessor type, then assign that question's id to the `togglePointsVisability` option.  The preselected option on the TImeDomainViewer must also NOT be set to true.


blnkt:
> Looks great! Works as expected. Only comment I may have is that once you save your selection/answer, the circled selection disappears.

could not reproduce.  Gonna rebase and merge.

PR #157, opened 2020-05-27
[EPO-3052] Add KECalculator

georgegomez144:
[F] Create the KECalculator

[EPO-3052]

[C] Create Kinetic Energy Calculator

[EPO-3052]
Fix storage bugs.
format user input and update.


src/components/qas/questions/qaCalculators/shared/equation/FindKineticEnergy.jsx
```diff
@@ -0,0 +1,40 @@
...
+import PropTypes from 'prop-types';
+import { addTheCommas, formatValue } from '../../../../../../lib/utilities';
+import {
+  equation,
+  fraction,
+  denominator,
+  exponent,
+} from './equation.module.scss';
+
+export default function FindKineticEnergy(props) {
```
blnkt, line 11: I really like the way the other calculators (for instance [the SizeCalc seen here](http://localhost:8000/identifying-pha-2/)) color (uses `$basePrimary`) the variables the user is providing values for (as well as the `?`).  Can we pull that styling over to this component as well?

src/components/qas/questions/qaCalculators/kineticEnergyCalculator/kineticEnergyCalculator.module.scss
```diff
@@ -0,0 +1,77 @@
+:global {
```
blnkt, line 1: again, I think this is a misuse of the global flag.  If it' use is unclear I'm happy to talk about it together or feel free to [spend some time](https://github.com/css-modules/css-modules/issues/264#issuecomment-609067219) reading up [about it](https://blog.yipl.com.np/css-modules-with-react-the-complete-guide-a98737f79c7c) We should try really hard to use this sparingly, ONLY for classes outside our control, and, if possible, scope it under a class we DO have control over (although I know that's not always gonna be possible).  I recognize there are many of those classes outside our control in this component, but you also have a lot of "our" classes inside the global flag too.  Those are gonna pollute our styles scope and, if used across like components, with like class names, will potentially cause unintended consequences.  I actually ran into just that problem recently and so cleaned up quite a bit of this kind of thing in the other calculator components.  Please look to those components for some indications of how to address the issue here in this file.

Additionally, I'm seeing a lot of unused classes in here.  Please delete anything that's not being used.  It's really important to eliminate unused code like this (and similar in javascript files) before we commit work to master because it visually clutters up our files, increases bloat, and can really hard to spot later on when this code is not fresh in our minds.

src/components/qas/questions/qaCalculators/kineticEnergyCalculator/index.jsx
```diff
@@ -0,0 +1,207 @@
...
+              defaultValue={answered ? mass : null}
+              onBlur={this.handleBlur}
+              onFocus={this.handleFocus}
+              onChange={value => this.handleChange(value, 'mass')}
+              disabled={!(answerable || answered || active)}
+            />
+          </div>
+          <div className={`col-50 ${styles.marginLeft}`}>
+            <TextField
+              id={`text-${isTextArea ? 'area' : 'input'}-${id}`}
```
blnkt, line 173: this can just be id={`text-input-${id}`} because you are dictating what style of input is used and it's never gonna be a textArea.  Actually it's never gonna be a text input it's a number input so you could call it that instead if you want

src/components/qas/questions/qaCalculators/kineticEnergyCalculator/index.jsx
```diff
@@ -0,0 +1,207 @@
...
+    return (
+      <Card
+        className={cardClasses}
+        expanded={hasFocus}
+        onExpanderClick={() => {}}
+      >
+        <CardText>
+          <div className={`col-50 ${styles.marginLeft}`}>
+            <TextField
+              id={`text-${isTextArea ? 'area' : 'input'}-${id}`}
```
blnkt, line 156: this can just be id={`text-input-${id}`} because you are dictating what style of input is used and it's never gonna be a textArea.  Actually it's never gonna be a text input it's a number input so you could call it that instead if you want

src/components/qas/questions/qaCalculators/kineticEnergyCalculator/index.jsx
```diff
@@ -0,0 +1,207 @@
...
+  render() {
+    const { question, answer, activeId } = this.props;
+    const {
+      hasFocus,
+      answerable,
+      value: { mass, velocity, kineticEnergy },
+    } = this.state;
+
+    const { id, questionType, label, placeholder } = question;
+    const isTextArea = questionType === 'textArea';
```
blnkt, line 135: this will never be true please delete

src/components/qas/questions/qaCalculators/kineticEnergyCalculator/index.jsx
```diff
@@ -0,0 +1,207 @@
...
+      answerable: answerable || answered || active,
+    });
+
+    return (
+      <Card
+        className={cardClasses}
+        expanded={hasFocus}
+        onExpanderClick={() => {}}
+      >
+        <CardText>
```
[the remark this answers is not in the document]
georgegomez144: I did something very similar in EPO-3161. I'm going to implement something similar into this solution.

src/components/qas/questions/qaCalculators/kineticEnergyCalculator/index.jsx
georgegomez144: updated to `text-number-${id}` per your suggestion.

src/components/qas/questions/qaCalculators/kineticEnergyCalculator/index.jsx
georgegomez144: updated to `number-input-${id}` per your suggestion.

src/components/qas/questions/qaCalculators/kineticEnergyCalculator/index.jsx
georgegomez144: updated to `text-input-${id}` per your suggestion.

PR #171, opened 2020-06-03
[EPO-3057] TimeDomainViewer selection populates OrbitViewer

blnkt:


src/components/qas/questions/qaExpansionList/styles.module.scss
```diff
@@ -57,7 +57,7 @@
   &.answered:not(.active) {
     :global {
       .md-panel-header {
-        padding: $minPadding;
+        padding: $minPadding $minPadding 0 $minPadding;
```
georgegomez144, line 60: Just a suggestion: you can remove the last `$minPadding`. CSS will render the second value as both `right` and `left`.

PR #172, opened 2020-06-04
[EPO-3053] Create generic compound qa type.

georgegomez144:
Add new generic component to render both text-input and select qa types
in the same compound qa.
Add this new 'questionType' check in the QASelect and QATextInput components.

To render this QA component you'll need to set the `questionType` to `compound`.
To render QASelect, set the `answerAccessor` in the question to `compound-select`.
To render QATextInput, set the `answerAccessor` in the question to `compound-input`.


blnkt:
closing without merging in favor of another branch with cleaner git history

PR #185, opened 2020-06-12
[C] Prevents Page Progression

parkerkaufmann:
Unanswered questions or empty answers prevent page progression.

### DIFF
--- a/src/components/pageNav/index.jsx
+++ b/src/components/pageNav/index.jsx
@@ -8,7 +8,7 @@ import ArrowRight from '../site/icons/ArrowRight';
 import styles from './page-nav.module.scss';
 
 class PageNav extends React.PureComponent {
-  renderNavItem(type, item, baseUrl) {
+  renderNavItem(type, item, baseUrl, disableButton = false) {
     const { link, title } = item;
     const linkIsBlank = link === '' || link === null;
     const isLinkToFirstPage = linkIsBlank && type === 'previous';
@@ -26,7 +26,7 @@ class PageNav extends React.PureComponent {
         icon
         className="outlined"
         to={buttonLink}
-        component={Link}
+        component={disableButton ? null : Link}
         iconEl={
           type === 'previous' ? (
             <ButtonIcon srText={title || 'Home'} Icon={ArrowLeft} />
@@ -34,6 +34,7 @@ class PageNav extends React.PureComponent {
             <ButtonIcon srText={title || 'Home'} Icon={ArrowRight} />
           )
         }
+        disabled={disableButton}
         iconBefore={type === 'previous'}
         tooltipLabel={item.title}
         tooltipPosition="top"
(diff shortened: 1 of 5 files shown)


src/containers/PageContainer.jsx
```diff
@@ -26,6 +26,22 @@ class PageContainer extends React.PureComponent {
     this.dispatch.updatePageId(id);
   }
 
+  shouldDisable = questions => {
+    const { answers } = this.props;
+    if (questions === undefined) {
+      return false;
+    }
```
blnkt, line 33: good on you for thinking of this case, however, `questions === null` when there are no questions on the page so this catch won't quite work as you would expect. try this instead: `if (!questions) return false;`

src/components/pageNav/index.jsx
```diff
@@ -8,7 +8,7 @@ import ArrowRight from '../site/icons/ArrowRight';
 import styles from './page-nav.module.scss';
 
 class PageNav extends React.PureComponent {
-  renderNavItem(type, item, baseUrl) {
+  renderNavItem(type, item, baseUrl, disableButton) {
```
blnkt, line 11: ok, so I think we can clean up some of this previous vs next button stuff.  First thing's first we'll provide a default value for `disableButton` so like: `renderNavItem(type, item, baseUrl, disableButton = false) {`.  That way, when we call this method for the previous button, we can make it so disableButton is always false...which should prove helpful

sidebar: this could also be affected by passing in `false` instead of `disableButton` when this method is called for the previous button, but I like the default args strat more for whatever reason

src/components/pageNav/index.jsx
```diff
@@ -26,14 +26,15 @@ class PageNav extends React.PureComponent {
         icon
         className="outlined"
         to={buttonLink}
-        component={Link}
+        component={type === 'previous' ? Link : disableButton ? null : Link}
```
blnkt, line 29: so, with the above change, is always false for previous button so we can simplify this to `component={disableButton ? null : Link}`

src/components/pageNav/index.jsx
```diff
@@ -26,14 +26,15 @@ class PageNav extends React.PureComponent {
...
         iconEl={
           type === 'previous' ? (
             <ButtonIcon srText={title || 'Home'} Icon={ArrowLeft} />
           ) : (
             <ButtonIcon srText={title || 'Home'} Icon={ArrowRight} />
           )
         }
+        disabled={type === 'previous' ? false : disableButton}
```
blnkt, line 37: so with the above change is always false for previous so we can simplify to `disabled={disableButton}`

src/components/pageNav/index.jsx
```diff
@@ -42,13 +43,14 @@ class PageNav extends React.PureComponent {
...
-    const { previous, next, baseUrl } = this.props;
+    const { previous, next, baseUrl, disableButton } = this.props;
 
     return (
       <div className={styles.pageNavigation}>
         <nav role="navigation" className={styles.navSecondary}>
-          {previous && this.renderNavItem('previous', previous, baseUrl)}
-          {next && this.renderNavItem('next', next, baseUrl)}
+          {previous &&
+            this.renderNavItem('previous', previous, baseUrl, disableButton)}
```
blnkt, line 52: see related changes; this becomes: `{previous && this.renderNavItem('previous', previous, baseUrl)}`

PR #192, opened 2020-06-23
[EPO-3038] Only update calc if needed

georgegomez144:


blnkt:
closing without merging because while complete, this work needs to be implemented similarly but in a different place after completing EPO-3236, EPO-3291, and EPO-3292

PR #198, opened 2020-06-30
[EPO-3278] AsteroidClass plot supports overlay line

blnkt:
Overlay line is used to present an single object to totality of objects' filters data


src/components/charts/asteroidClass/asteroid-class.module.scss
```diff
@@ -13,19 +13,46 @@
...
+}
+
 .x-axis-label {
   text-anchor: middle;
 }
 
 .point {
   opacity: 0.2;
-  fill: $mediumBlue;
+  fill: $chart3;
```
georgegomez144, line 30: More like a heads up, I removed $chart4 from the colors as I was not able to find and acceptable 'yellow' that was passing the contrast test.

PR #199, opened 2020-06-30
[EPO-3277] AsteroidClass plot supports multiple datasets

blnkt:
Utilizes NavDrawer pattern to support toggling between multiple datasets (in this case toggling overlaid object filter data);


src/components/charts/asteroidClass/FilterValuesLine.jsx
```diff
@@ -18,7 +18,18 @@ class Line extends React.PureComponent {
       <g>
         {filters.map((filter, i) => {
           const nextFilter = filters[i + 1];
-          if (!nextFilter) return null;
+          if (!nextFilter) {
+            return (
+              <g key={`point-${filter}-${nextFilter}`}>
```
georgegomez144, line 23: wouldn't `nextFilter` here just render out `undefined`?

PR #214, opened 2020-07-13
[EPO-3291] Consolidate QACalcs

georgegomez144:
- Equations live together once again.
- Broke out all the common functions into one file: index.jsx
- All custom calc data sits inside qaCalculatorsData.jsx
- Updated units (and equation conversions) to use <Unit /> component


src/components/qas/questions/qaCalculator/index.jsx
```diff
@@ -1,66 +1,246 @@
...
+      [activeClass]: hasFocus,
     });
+    const fieldClasses = classnames('qa-text-input', answeredClasses);
+    const labelClasses = classnames(calcLabel, answeredClasses);
+
+    const { inputs, equation } = this.calculator[questionType];
+    const FindEquation = equation || null;
 
-    const QACalc = this.calculators[questionType];
+    const isDistCalc = includes(questionType, 'DistanceCalculator');
```
blnkt, line 174: Not sure `includes` is necessary.  Should probably just be a normal comparison of strings if possible.  Additionally I'll point out that the need for this `isDistCalc` variable might bely one of the shortcomings of this approach.  For instance, what happens if there's another calculator with an expanding bit?

src/components/qas/questions/qaCalculator/qaCalculatorIcons.jsx
```diff
@@ -21,10 +21,9 @@ QACalculatorIcon.propTypes = {
 
 export const QACalculatorIconUnit = ({ unit, tiny, className }) => {
   return (
-    <span
-      className={classnames(icon, className, { [smallIcon]: !!tiny })}
-      dangerouslySetInnerHTML={renderDef(unit)}
-    />
+    <span className={classnames(icon, className, { [smallIcon]: !!tiny })}>
```
blnkt, line 24: If tiny is false by default I believe this can just be `classnames(icon, className, { [smallIcon]: tiny })` since even when not provided this prop would always be `true` or `false` right?

src/components/qas/questions/qaCalculator/qaCalculatorIcons.jsx
```diff
@@ -21,10 +21,9 @@ QACalculatorIcon.propTypes = {
 
 export const QACalculatorIconUnit = ({ unit, tiny, className }) => {
```
blnkt, line 22: Typically we've been writing React component filenames in CapitalCamelCase and reserving lowercaseCamelCase for vanilla js files (index.jsx files not withstanding).

src/components/qas/questions/qaCalculator/qaCalculatorsData.jsx
```diff
@@ -0,0 +1,120 @@
...
+import FindKineticEnergy from './equations/FindKineticEnergy';
+import {
+  getCalculatedMeasurementsForDistance,
+  calculateDiameter,
+  calculateKineticEnergy,
+  calculateCraterDiameter,
+} from './qaCalculatorUtilities';
+import { col50 } from './qaCalculator.module.scss';
+import Unit from '../../../charts/shared/unit';
+
```
blnkt, line 14: You're right.  This is not what I had in mind, but it certainly does a great job of consolidating the work and organizing information unique to each calculator in a manner that's both fairly readable, and fairly easy to update info in the future.  I think it's perhaps debatable whether this approach allows for greater flexibility.  And I think the Distance Calculator is a good example of how this: you had to create a condition outside of this system of calc definitions, that uses a hardcoded string value to conditionally render a pretty sizable chunk of mark up, a layout/style, I might add, I may already be anticipating using in the Impact Calculator.  Additionally, with this approach, if you want/need to change the name of a prop, you'll not only have to update it wherever that prop is used but also across this file as well.  I'm not suggesting this is a bad approach, I just want to recognize that it's not without its complications.

I think what I would have done is used the framework of react components as the organizational system.  So, rather than have a js object that represents the info unique to a calc, you would have had a React component for each calc that contains all the unique markup, styles, logic etc.  Any styles, logic, or markup not unique to the calc can be passed into it as props (in the traditional way or by way of a Higher Order Component), or live in a parent component itself (as you have done).  That would mean things like the active states, answerable, all the event handlers, some styles, some markup, etc.  And anything unique to a calc, like its form fields, markup, the math, how answers are presented, lives in its own component.  And in some cases, as you have done, even some of those unique things can be abstracted into their own reusable components (if sufficiently similar).  

I'm not at all suggesting you revisit this work to affect "how I would have done it," but I did want to share my perspective, and invite you to discuss it further if I have not well described my competing approach.

src/components/qas/questions/qaCalculator/qaCalculatorsData.jsx
```diff
@@ -0,0 +1,120 @@
...
+    },
+  ],
+};
+
+export const ImpactCalculator = {
+  value: {
+    craterDiameter: null,
+    density: null,
+    velocity: null,
+    asteroidDiameter: 156,
```
blnkt, line 92: typo?

src/components/qas/questions/qaCalculator/qaCalculatorIcons.jsx
```diff
@@ -33,7 +32,7 @@ QACalculatorIconUnit.defaultProps = {
 };
 
 QACalculatorIconUnit.propTypes = {
-  unit: PropTypes.string,
+  unit: PropTypes.oneOfType(PropTypes.string, PropTypes.object),
```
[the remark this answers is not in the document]
georgegomez144: I moved the `<Unit />` into the qaCalculatorIcons component and updated the code as needed. No longer need the `.oneOfType` prop.

src/components/qas/questions/qaCalculator/index.jsx
georgegomez144: I agree, as this was a thought of mine while I was updating this. I'm still not certain how I'd like to handle this going forward and thought that if the case ever came up I might have a better mind to solve this.

src/components/qas/questions/qaCalculator/qaCalculatorIcons.jsx
georgegomez144: solid catch here! Thank you!

src/components/qas/questions/qaCalculator/qaCalculatorIcons.jsx
georgegomez144: Updated 👍

src/components/qas/questions/qaCalculator/qaCalculatorsData.jsx
georgegomez144: yeah, it was apart of my testing...forgot about it. It has been updated now. 👍

src/components/qas/questions/qaCalculator/qaCalculatorsData.jsx
georgegomez144: I would love to see your perspective on this. I was very much stuck, in a way, that I was very uncomfortable with "how to do it the react way". The approach was not clear to me and I'm 100% certain that this could become even **MORE**, certainly **better and efficient**, if approached from the React view point. Something to discuss in our 2pm Thursday meeting.

src/components/qas/questions/qaCalculator/qaCalculatorIcons.jsx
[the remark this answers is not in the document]
blnkt: Oh ok I think maybe I understand you.  If it was the case that you were trying to set it up such that `unit` could be a string or an instance of `</Unit />` that doesn't sound unreasonable.  But even if that should be the case you should use either `node`, `element`, or even `instanceOf(Unit)`.  If that was your initial intent...

PR #228, opened 2020-07-24
[F] Control Panel Overlay

parkerkaufmann:


src/components/charts/colorMixingTool/index.jsx
```diff
@@ -232,24 +227,35 @@ class ColorTool extends React.PureComponent {
...
+            Reset
+          </Button>
         </div>
+        <Button
+          floating
+          className={panelClassName}
+          onClick={this.handleControlPanel}
+        >
+          {panelIsActive ? 'Off' : 'On'}
+        </Button>
```
georgegomez144, line 247: some css to help center your text:
```
padding: 0;
text-align: center;
```
You can apply these directly to the element via `style={{}}` or use a class. I would most likely suggest a class in this instance.

src/assets/stylesheets/_variables.scss
```diff
@@ -63,9 +63,13 @@ $chartColors: $chart0 $chart1 $chart2 $chart3 $chart4 $chart5 $chart6;
 $sliderTrackFillRed: red;
 $sliderTrackFillBlue: blue;
 $sliderTrackFillGreen: green;
+$sliderTrackFillPurple: violet;
+$sliderTrackFillCyan: cyan;
+$sliderTrackFillYellow: yellow;
+$sliderTrackFillOrange: orange;
```
georgegomez144, line 69: There might be a discussion later about the specific hex codes that will be used, but this is certainly great start and personally, I think these do just fine 👍

src/components/charts/colorMixingTool/color-tool.module.scss
```diff
@@ -48,7 +48,6 @@
 
 .images-container {
   position: relative;
-  height: 300px;
   padding: $minPadding;
   overflow: hidden;
 }
```
georgegomez144, line 53: There is some cool aspect-ratio stuff you can do with the code to keep a specific height regardless of width. It looks something like this:
```
.your-class-name {
  width: 100%;
  padding-top: 100%; // for 1:1 ratio
}
```
You can find some doc stuffs about [aspect ratio here](https://www.w3schools.com/howto/howto_css_aspect_ratio.asp)

PR #230, opened 2020-07-30
[EPO-3408] Added additional calculators and textual outputs

georgegomez144:
Add crater depth calcuator (values are calculating incorrectly)
Add richtor magnitude calculator
Add seismic damage textual output
Add air blast calculator (needs much more work)
Add textual output for air blast damage
Create ticket for the additional work not finish in this ticket.

Work finalized in this ticket.

### DIFF
--- a/src/components/qas/questions/qaCalculator/equations/FindImpactCrater.jsx
+++ b/src/components/qas/questions/qaCalculator/equations/FindImpactCrater.jsx
@@ -1,5 +1,8 @@
+/* eslint-disable react/no-array-index-key */
+/* eslint-disable react/no-danger */
 import React from 'react';
 import PropTypes from 'prop-types';
+import classnames from 'classnames';
 import { addTheCommas, formatValue } from '../../../../../lib/utilities.js';
 import { QACalculatorIcon, QACalculatorIconUnit } from '../qQaCalculatorIcons';
 import {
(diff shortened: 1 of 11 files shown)


src/components/qas/questions/qaCalculator/index.jsx
```diff
@@ -197,13 +212,18 @@ class QACalculator extends React.PureComponent {
                     <TextField
                       id={`text-input-${id}-${defaultValue}`}
                       data-testid={`qa-calc-input-${i}`}
-                      className={fieldClasses}
+                      data-default-value={defaultValue}
```
blnkt, line 215: I'm not sure what this prop is doing?

src/components/qas/questions/qaCalculator/index.jsx
```diff
@@ -197,13 +212,18 @@ class QACalculator extends React.PureComponent {
...
+                      })}
                       leftIcon={leftIcon}
                       rightIcon={rightIcon}
+                      label={inputLabel}
                       placeholder={placeholder}
                       lineDirection="center"
                       type="number"
-                      min="0"
+                      min={min}
+                      max={max}
```
blnkt, line 226: so good news/bad news.  The good news is you [preempted a comment I was going to make](https://jira.lsstcorp.org/browse/EPO-3501) regarding allowing a range of values as Lauren suggested: .  Excellent.  Bad News is these min/max props are not doing anything because a number input only cares about the min and max when you're iterating the number using the lil up and down arrows, or if the input is submitted as part of a form.  We never use a form element, and there is no submit, so the min and max are never honored.  

Now, we do control the onChange handler, and we can control the value attribute, so we could make this a "controlled" input and screen values against the min and max ourselves, but I have done that before on other projects and it always gets messy.  So, instead, I suggest we leave this as-is and I'm gonna backlog the ticket I made until we have a better idea about it.

src/components/qas/questions/qaCalculator/qaCalculatorUtilities.js
```diff
@@ -51,6 +51,63 @@ export const calculateKineticEnergy = ({ mass, velocity }) =>
 
 export const calculateMass = ({ density, volume }) => density * volume;
 
+export const calculateDestruction = () => {};
+
+export const calculateMinimumSafe = () => {};
+
```
blnkt, line 57: yes, you're right, we'll have to figure out what constitutes these definitions next.  Creating a ticket to update the Impact Calc with this info.  I think is as straightforward as us just deciding based on the damage descriptions what we (i.e. Lauren and Ardis) agree to be the "Area of Destruction" and "Minimum Safe Distance".

src/components/qas/questions/qaCalculator/qaCalculatorUtilities.js
```diff
@@ -64,5 +121,129 @@ export const calculateCraterDiameter = ({
...
+    '<em style="font-weight: 500;">IX.</em> General panic. Damage considerable in specially designed structures; well-designed frame structures thrown out of plumb. Damage great in substantial buildings, with partial collapse. Buildings shifted off foundations. Serious damage to reservoirs. Underground pipes broken. Conspicuous cracks in ground. In alluviated areas sand and mud ejected, earthquake fountains, sand craters.',
+    '<em style="font-weight: 500;">X.</em> Most masonry and frame structures destroyed with their foundations. Some well-built wooden structures and bridges destroyed. Serious damage to dams, dikes, and embankments. Large landslides. Water thrown on banks of canals, rivers, lakes, etc. Sand and mud shifted horizontally on beaches and flat land. Rails bent slightly.',
+    '<em style="font-weight: 500;">XI.</em> As X. Rails bent greatly. Underground pipelines completely out of service.',
+    '<em style="font-weight: 500;">XII.</em> As X. Damage nearly total. Large rock masses displaced. Lines of sight and level distorted. Objects thrown into the air.',
+  ];
+
+  return mercalliIntensity[Math.round(richterMagnitude)];
+};
```
blnkt, line 150: It's usually bad form to throw blocks of stringified html into an array like this.  I'd like to see this take a format more akin to the Air Blast damage descriptions.  Furthermore, I think that's pretty doable as I don't think the Mercalli Intensities are useful here.  A lot/everybody knows about the richter scale, so I thought it would be fun to include that number, but no one has ever heard of this other scale.  If you aren't having to bold the roman numerals are you then able to treat these descriptions the same way you did the air blast ones?

"As X." won't mean anything for the user since we are not (by design currently) giving them access to the with the full table of descriptions, so let's go ahead and replace both of those with the full description of "X": "Most masonry and frame structures destroyed with their foundations. Some well-built wooden structures and bridges destroyed. Serious damage to dams, dikes, and embankments. Large landslides. Water thrown on banks of canals, rivers, lakes, etc. Sand and mud shifted horizontally on beaches and flat land. Rails bent slightly" (and should be tweaked where appropriate to avoid redundant language).

And one more note here: I think your approach here (clever rounding + indexing + look up of the richter value) is probably why richter values are not lining up as they are in the the table (as notes in the main review comment).  I believe this approach would work if the descriptions lined up 1:1 with richter values, but there are jumps in terms of what ranges correspond to which ranges of intensities, so I think you are going to have to use an approach similar to what you're doing with the Air Blast look up.  Otherwise I do agree that if a richter rounds up or down only show that one corresponding mercalli description.

src/components/qas/questions/qaCalculator/tests/ImpactCalculator.test.js
```diff
@@ -21,35 +21,53 @@ test('ImpactCalculator renders with required props', () => {
...
   await waitFor(() => {
     expect(input1.value).toBe('1');
     expect(input2.value).toBe('3000');
     expect(input3.value).toBe('17');
-    expect(equationKE).toHaveTextContent('10.577 km');
+    expect(input4.value).toBe('200');
+    expect(equationImpact).toHaveTextContent('10.577 km');
+    expect(equationImpact).toHaveTextContent('3.74 km');
+    expect(equationImpact).toHaveTextContent(
+      'Richtor Magnitude at 200 km from impact: 5.6'
```
blnkt, line 60: Richter not Richtor.

src/components/qas/questions/qaCalculator/equations/FindImpactCrater.jsx
```diff
@@ -1,85 +1,180 @@
...
+            </span>
+          </p>
+        </div>
+      )}
+      {richterMagnitude && (
+        <div className={topSpacing}>
+          <hr />
+          <p>
+            <span className={boldText}>
+              Richtor Magnitude at {addTheCommas(observerDistance)}{' '}
```
blnkt, line 129: Richter not Richtor.

src/components/qas/questions/qaCalculator/index.jsx
georgegomez144: Yeah, I had something for this inside the onBlur handler and well, it wasn't updated after the input was changed...I wanted to look into it more, but decided not to at this time. Certain worth the visit down the road. There was much more work on this than you see now that i've ripped out.

src/components/qas/questions/qaCalculator/index.jsx
georgegomez144: I was using this for the min/max bits

src/components/qas/questions/qaCalculator/qaCalculatorUtilities.js
georgegomez144: calculating those values current eludes me, at the moment, but I'm sure as we approach this work it will become more clear.

src/components/qas/questions/qaCalculator/qaCalculatorUtilities.js
```diff
@@ -64,5 +122,180 @@ export const calculateCraterDiameter = ({
...
+    const { overPressure } = damage;
+    return overPressure <= airBlastOverPressure;
+  });
+
+  return filteredAirBlastDamage;
+};
+
+export const calculateAsteroidImpact = props => {
+  // console.log({ ...props });
+  const craterDiameter = calculateCraterDiameter(props);
```
blnkt, line 278: This is a neat trick with feeding in `props` and then spreading on the other side.  We don't make use of this trick too often (outside the obvious use in writing React components), and the reason for that is it can make it kinda challenging to trace relationships/interdependencies between methods.  I think your use here is shy of that line and A-ok, but I wouldn't recommend using this as a go to strat unless the method your giving props to spreads and then uses most of the props.  If you're spreading just to grab one or two things...well, maybe not totally justifiable.

src/components/qas/questions/qaCalculator/qaCalculatorsData.jsx
```diff
@@ -7,7 +7,7 @@ import {
   getCalculatedMeasurementsForDistance,
   calculateDiameter,
   calculateKineticEnergy,
-  calculateCraterDiameter,
+  calculateAsteroidImpact,
```
blnkt, line 10: Good call

src/components/qas/questions/qaCalculator/qaCalculatorUtilities.js
georgegomez144: Okay. I believe I did this just to get going. Felt smooth enough, but I do understand your point. I went ahead and updated all them to be explicit.

src/components/qas/questions/qaCalculator/qaCalculatorUtilities.js
```diff
@@ -64,5 +122,180 @@ export const calculateCraterDiameter = ({
...
+    {
+      id: 'XII',
+      range: [9, '+'],
+      description:
+        'As X. Damage nearly total. Large rock masses displaced. Lines of sight and level distorted. Objects thrown into the air.',
+    },
+  ];
+
+  return mercalliIntensity.filter(mi =>
+    includes(mi.range, Math.round(richterMagnitude))
```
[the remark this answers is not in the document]
georgegomez144: You were correct on this one. `inRange` was a much better function in this case.

src/components/qas/questions/qaCalculator/qaCalculatorUtilities.js
```diff
@@ -64,5 +120,191 @@ export const calculateCraterDiameter = ({
...
+    },
+    {
+      id: 'XI',
+      range: [8.5, 9],
+      description:
+        'Rails bent greatly. Underground pipelines completely out of service.',
+    },
+    {
+      id: 'XII',
+      range: [9, 20],
```
blnkt, line 202: If this is an arbitrary cut off at 20 let's go ahead and replace 20 with `Infinity` so the user always gets a description regardless of what huge richter value they manufacture.  If this 20 is an intentional hard cap then ignore this coment.

PR #231, opened 2020-07-30
[F] Add page numbers to header and table of contents

georgegomez144:
Used the pages' index (+ 1) and stored that in the pages array as 'pageNumber' before storing the pages into global store.

Used that pages array in the Table of contents simply by adding the new pageNumber index in the primaryText.

Pushed a filtered array of pages (only page.id and page.pageNumber) into the Header component. Used the 'lodash/find' to get the page from my new pages array and set that to a pageNumber variable that outputs ": Page {pageNumber}". Imported globalHistory from '@reach/router' and used it's location.pathname prop to catch the page url for the 'last-page' instance as well as the introduction page.

### DIFF
--- a/src/components/site/header/header.test.js
+++ b/src/components/site/header/header.test.js
@@ -14,11 +14,19 @@ test('Header renders without props', () => {
 
 test('Header Logo alt text is Investigation Title', () => {
   // Arrange
-  const { getByAltText, getByText } = render(<Header logo="fake/image/path" investigationTitle={testTitle} />);
+  const { getByAltText, getByText, queryByText } = render(<Header logo="fake/image/path" investigationTitle={testTitle} />);
   const logoLink = getByAltText(testTitle);
   // Assert
   expect(getByText(testTitle)).toBeInTheDocument();
+  expect(queryByText(`${testTitle}: Page`)).toBe(null);
   expect(logoLink).toBeInTheDocument();
+});
+
+test('Header renders page number if provided', () => {
+  // Arrange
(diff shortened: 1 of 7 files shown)


src/components/site/header/header.test.js
```diff
@@ -14,13 +14,21 @@ test('Header renders without props', () => {
 
 test('Header Logo alt text is Investigation Title', () => {
   // Arrange
-  const { getByAltText, getByText } = render(<Header logo="fake/image/path" investigationTitle={testTitle} />);
+  const { getByAltText, getByText, queryByText } = render(<Header logo="fake/image/path" investigationTitle={testTitle} />);
   const logoLink = getByAltText(testTitle);
   // Assert
   expect(getByText(testTitle)).toBeInTheDocument();
+  expect(queryByText(`${testTitle}: Page`)).toBe(null);
```
georgegomez144, line 21: Great solution

src/layouts/index.jsx
```diff
@@ -129,11 +124,10 @@ class Layout extends React.Component {
       <>
         <SEO title={investigation || 'Investigation'} />
         <Header
-          investigationTitle={investigationTitle}
+          {...{ investigationTitle, logo }}
+          pageNumber={this.getCurrentPageNumber()}
```
georgegomez144, line 128: I like this method much better.

src/state/GlobalStore.js
```diff
@@ -56,6 +56,13 @@ class GlobalStore {
         visitedPages: prevVisitedPages,
       } = global;
 
+      if (!pageId) {
+        return {
+          ...global,
+          pageId,
+        };
+      }
+
```
georgegomez144, line 65: this feels much better

PR #235, opened 2020-08-12
[EPO-3416] Add question numbers to investigations

georgegomez144:
When getting total qas on page, set questions numbers by page
Use stored qa number in global store to populate question number
Format and catch all breaking layouts within labels to have number sit
  next to first line in label

### DIFF
--- a/src/components/qas/styles.module.scss
+++ b/src/components/qas/styles.module.scss
@@ -13,3 +13,9 @@
     box-shadow: none;
   }
 }
+
+.label-with-number {
+  p:first-child() {
+    display: inline;
+  }
+}
--- a/src/containers/PageContainer.jsx
+++ b/src/containers/PageContainer.jsx
@@ -17,11 +17,22 @@ class PageContainer extends React.PureComponent {
       TwoCol,
       SingleCol,
     };
+
+    this.state = {
+      questions: null,
+    };
   }
 
   componentDidMount() {
     const { data } = this.props;
-    const { id } = data.allPagesJson.nodes[0];
+    const { id, questionsByPage: questions } = data.allPagesJson.nodes[0];
+    const { questionNumbersByPage } = this.global || {};
+    const questionNumbers = questionNumbersByPage[id];
+
+    this.setState(prevState => ({
+      ...prevState,
+      questions: this.getQuestionsWithNumbers(questionNumbers, questions),
+    }));
 
     this.dispatch.updatePageId(id);
   }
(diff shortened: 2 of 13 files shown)


src/state/GlobalStore.js
```diff
@@ -19,6 +19,7 @@ class GlobalStore {
       visitedPages: [],
       totalQAsByInvestigation: null,
       totalQAsByPage: null,
+      questionNumbersByPage: null,
```
blnkt, line 22: So this was the big thing.  You had this guy sitting out here going unused and it got me to thinking, yea, this is definitely where these page numbers should live, and they should be indexed by page Id, let's do that (instead of your kinda shoehorning the question numbers in amongst the totals and qa/page progress, which was also indexed by page id, so was a super convenient place to store this question numbering stuff, but it didn't really "belong").

src/layouts/index.jsx
```diff
@@ -66,6 +68,29 @@ class Layout extends React.Component {
...
+          return qNum;
+        });
+      } else {
+        qsByPage[id] = [];
+      }
+    });
+
+    return qsByPage;
+  }
+
```
blnkt, line 91: So here I updated your getPageNumbers method to do basically the same thing, just as a loop within the loop so we could get the question numbers for every page and then spit out an object where the page ids are the keys and the values are arrays of question numbers (or empty arrays).

src/containers/PageContainer.jsx
```diff
@@ -96,9 +118,9 @@ class PageContainer extends React.PureComponent {
             widgets,
             tables,
             images,
-            questions,
             answers,
             shared,
+            questions,
```
blnkt, line 123: this is another sticking point.  So you were running your getQuestionNumbers method inline here.  It's important to remember that any methods in the render method run every time the render method runs.  In this case that's quite often because it is responding to changes on the page like new/modified answers, and even questions getting their active state applied, and all manner of other stuff.  We couldn't expect our question numbers to change, so we should set once and forget about it.  So I added questions to this component's local state and do just that in the componentDidMount.

src/containers/PageContainer.jsx
```diff
@@ -46,6 +57,16 @@ class PageContainer extends React.PureComponent {
...
+  getQuestionsWithNumbers = (questionNumbers, questions) => {
+    if (!questions || !questionNumbers) return [];
+
+    questions.forEach((question, index) => {
+      question.number = questionNumbers[index];
+    });
+
+    return questions;
+  };
+
```
blnkt, line 69: basically what you had, just using it in a bit more straight forward way now that  we have this discreet object where we have arrays of question numbers organized by page id.

src/components/qas/styles.module.scss
```diff
@@ -13,3 +13,9 @@
     box-shadow: none;
   }
 }
+
+.label-with-number {
+  p:first-child() {
```
blnkt, line 18: don;'t think we need the `:global` in these cases because there is no global classname on our p tags to override...and I mostly tested this hypothesis...

PR #236, opened 2020-08-14
[EPO-3502] Update AirBlast descriptions

georgegomez144:
Organize descriptions by category within their damage ranges
Use Lodash's 'find' function to pull out first found in array
Push all responses into array and print them on screen


blnkt:
Good work!  Thank you for re-organizing this set of info to be more succinct and to the point.

PR #242, opened 2020-08-27
Saving ColorTool Data

parkerkaufmann:


src/containers/AstroToolContainer.jsx
```diff
@@ -1,48 +1,45 @@
...
+    const { questionId } = options || {};
+    const { answers } = this.props;
+    const answer = answers[questionId];
     API.get(source).then(response => {
       const { data } = response;
       this.setState(prevState => ({
         ...prevState,
         data,
         options,
+        selectedFilters: answer === undefined ? null : answer.data,
```
georgegomez144, line 28: This would be best written as:
```
...
selectedFilters: answer ? answer.data : null;
```

src/containers/AstroToolContainer.jsx
```diff
@@ -54,8 +51,8 @@ class AstroToolContainer extends React.PureComponent {
     const { data, options, selectedFilters } = this.state;
     const { filters, colorOptions, hexColors, galaxies, galaxyImgs } =
       data || {};
-    const { galaxyImg } = options || {};
-    console.log(selectedFilters);
+    const { galaxyImg, saveAnswer } = options || {};
```
georgegomez144, line 54: I'm thinking that instead of using `saveAnswer` as your flag, you can similarly us `questionId` and have that be your flag to trigger in the Color Tool whether or not to save the data.

src/data/pages/astronomical-color-image-1.json
```diff
@@ -22,7 +22,8 @@
       },
     "source":  "/data/filters/astro-filter.json",
     "options": {
-      "questionId": "21"
+      "questionId": "21",
+      "saveAnswer": true
```
georgegomez144, line 26: Which means you won't need to have the `saveAnswer` here any longer 👍

src/fragments/widgetOptions.js
```diff
@@ -46,5 +46,7 @@ export const widgetOptionsFragemnt = graphql`
     required
     defaultZoom
     galaxyImg
+    questionId
+    saveAnswer
```
georgegomez144, line 50: And you won't need to add `saveAnswer` here since you can use `questionId` as your flag.

PR #252, opened 2020-09-11
[EPO-3693] Size & Impact Calcs use meters NOT kms

blnkt:
Uses meters presentationally and for all applicable math;
Math dependent on unit specific constants and operations use properly converted values (i.e. honor existing units);
Fixes some minor rounding problems;

### DIFF
--- a/src/components/qas/questions/qaCalculator/equations/FindImpactCrater.jsx
+++ b/src/components/qas/questions/qaCalculator/equations/FindImpactCrater.jsx
@@ -3,7 +3,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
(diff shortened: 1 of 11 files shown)


src/components/qas/questions/qaCalculator/equations/FindImpactCrater.jsx
```diff
@@ -119,7 +123,7 @@ export default function FindImpactCrater(props) {
           </p>
         </div>
       )}
-      {richterMagnitudeAtObserverDistance && (
+      {richterMagnitudeAtObserverDistance > 0 && (
```
georgegomez144, line 126: Nice solution here 👍

georgegomez144 (approved):

PR #274, opened 2020-09-28
[EPO-3479] Add css class to handle layout fix

georgegomez144:


blnkt:
This work does not address the issue at hand. what we would like to avoid is the layout break as seen in the attached screenshot.  I'm sorry if that was unclear in the ticket description.  If it's still unclear what the desired behavior is, shoot me a message and we can talk through it together.[](url)
Screen Shot 2020-09-28 at 12 10 00 PM

blnkt (changes requested):
This work does not address the issue at hand. what we would like to avoid is the layout break as seen in the attached screenshot.  I'm sorry if that was unclear in the ticket description.  If it's still unclear what the desired behavior is, shoot me a message and we can talk through it together.[](url)
Screen Shot 2020-09-28 at 12 10 00 PM

georgegomez144:
The update keeps the widget blinker from scaling past the toolbar/header. I adapted the layout to `overflow-x: auto` in the horizontal layout to help with that.
image

blnkt (approved):

PR #278, opened 2020-09-29
[EPO-3782] Fix introduction header increase issue

georgegomez144:


blnkt:
Good catch.  I went ahead and implemented this change across all investigations since it's heading that's in all of the investigations.  And I went ahead and included that work in #279 since it wasn't merged yet and already had a bunch of copy updates.  So gonna close this PR without merging

PR #280, opened 2020-10-01
[F] Added Loading Spinner to Histogram

parkerkaufmann:


blnkt:
excellent work @parkerkaufmann.  I tweaked the styles you had a smidge both to get the loading spinner to be the same size as we see it elsewhere in the investigations, and also to make the wrapper dive a nice responsive square for it to live in.

PR #287, opened 2020-10-15
[B] Replaced PaddingDrawerInner for useless main content class

parkerkaufmann:
Fixed overlap problem on light curve and found that other files were using an undefined class.
EPO-3409

### DIFF
--- a/src/containers/AsteroidClassContainer.jsx
+++ b/src/containers/AsteroidClassContainer.jsx
@@ -9,7 +9,6 @@ import AsteroidClass from '../components/charts/asteroidClass/index.jsx';
 
 import {
   drawerContainer,
-  mainContent,
   paddedDrawerInner,
   navItem,
   avatarContainer,
@@ -197,10 +196,10 @@ class AsteroidClassContainer extends React.PureComponent {
           <NavDrawer
             cardClasses={drawerContainer}
             navItems={this.generateNavItemsMulti(overlayData)}
-            contentClasses={mainContent}
+            contentClasses={paddedDrawerInner}
             toolbarStyles={{ display: 'none' }}
           >
-            <div className={paddedDrawerInner}>{children}</div>
+            {children}
           </NavDrawer>
         )}
       >
@@ -240,10 +239,10 @@ class AsteroidClassContainer extends React.PureComponent {
               <NavDrawer
                 cardClasses={drawerContainer}
                 navItems={this.generateNavItems(groups)}
-                contentClasses={mainContent}
+                contentClasses={paddedDrawerInner}
                 toolbarStyles={{ display: 'none' }}
               >
-                <div className={paddedDrawerInner}>{children}</div>
+                {children}
               </NavDrawer>
             )}
           >
--- a/src/containers/AsteroidClassTwoUpContainer.jsx
+++ b/src/containers/AsteroidClassTwoUpContainer.jsx
@@ -9,7 +9,6 @@ import AsteroidClass from '../components/charts/asteroidClass/index.jsx';
 
 import {
   drawerContainer,
-  mainContent,
   paddedDrawerInner,
   navItem,
   avatarContainer,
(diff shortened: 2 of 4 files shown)


src/components/charts/lightCurve/light-curve.module.scss
```diff
@@ -9,6 +9,13 @@
   }
 }
 
+.padded-drawer-inner {
+  padding-right: $minPadding;
+  padding-bottom: $minPadding;
+  padding-left: 72px;
+  outline: none;
+}
+
```
blnkt, line 18: Unless there's some conflict, I'd like to use the `padded-drawer-inner` class in `charts/shared/navDrawer/nav-drawer.module.scss` instead of duplicating the styles here.  Since that's like a "shared" style we're applying in a lot of the implementations of this NavDrawer.

src/containers/AsteroidClassContainer.jsx
```diff
@@ -193,16 +192,14 @@ class AsteroidClassContainer extends React.PureComponent {
     return (
       <ConditionalWrapper
         condition={multiple && !!overlayData}
-        wrapper={children => (
```
blnkt, line 196: Ah.  I understand your intuition here, but this is not quite right.  We actually still need to do this `children` magic in order for this ContionalWrapper component to work right.  It'll make more sense as you begin to work on Please revert to how it was previous [EPO-3932](https://jira.lsstcorp.org/browse/EPO-3932) next sprint (or you can do that ticket this sprint if you prefer).  Basically `children` is this like variable standing in for whatever CondiitonalWrapper is conditionally wrapping.  And the `wrapper` prop is maybe a bit misleadingly named as it's not JUST the component you want to be the wrapper, but is a function that returns the whole kit-and-kaboodle: the wrapper component and the children (the stuff being wrapped).  I'm confusing myself as I'm writing it.  Like I said we can chat about it more when you start working on EPO-3932.  For not change this line back to:
`wrapper={children => (`

and I'll indicate below how to change the next bit

src/containers/AsteroidClassContainer.jsx
```diff
@@ -193,16 +192,14 @@ class AsteroidClassContainer extends React.PureComponent {
...
+        wrapper={
           <NavDrawer
             cardClasses={drawerContainer}
             navItems={this.generateNavItemsMulti(overlayData)}
-            contentClasses={mainContent}
+            contentClasses={paddedDrawerInner}
```
blnkt, line 199: This part is right, but the next part...

src/containers/AsteroidClassContainer.jsx
```diff
@@ -193,16 +192,14 @@ class AsteroidClassContainer extends React.PureComponent {
...
             toolbarStyles={{ display: 'none' }}
-          >
-            <div className={paddedDrawerInner}>{children}</div>
-          </NavDrawer>
-        )}
+          />
+        }
```
blnkt, line 202: ...should be:
```
          >
            {children}
          </NavDrawer>
        )}
```

src/containers/AsteroidClassTwoUpContainer.jsx
```diff
@@ -133,16 +132,14 @@ class AsteroidClassTwoUpContainer extends React.PureComponent {
...
             navItems={this.generateNavItems(overlayData)}
-            contentClasses={mainContent}
+            contentClasses={paddedDrawerInner}
             toolbarStyles={{ display: 'none' }}
-          >
-            <div className={paddedDrawerInner}>{children}</div>
-          </NavDrawer>
-        )}
+          />
+        }
```
blnkt, line 142: see my comments above in `AsteroidClassContainer` and do same here

src/containers/OrbitalViewerContainer.jsx
```diff
@@ -167,17 +166,15 @@ class OrbitalViewerContainer extends React.PureComponent {
...
               cardClasses={container}
               navItems={this.generateNavItems(data)}
-              contentClasses={mainContent}
-            >
-              <div className={paddedDrawerInner}>{children}</div>
-            </NavDrawer>
-          )}
+              contentClasses={paddedDrawerInner}
+            />
+          }
```
blnkt, line 177: see my comments above in AsteroidClassContainer and do same here

blnkt (changes requested):
close, but not quite right way to use this ConditionalWrapper component.  But I don't think you've seen a so called "Higher Order Component" like this, so no biggie.  Happy to chat about it more, especially if my comments here aren't clear.

PR #292, opened 2020-11-10
[F] Replaced ShowNavDrawer with  Conditional Wrappers

parkerkaufmann:
EPO-3932

### DIFF
--- a/src/components/charts/shared/navDrawer/index.jsx
+++ b/src/components/charts/shared/navDrawer/index.jsx
@@ -44,7 +44,6 @@ class NavDrawer extends React.PureComponent {
       toolbarTitle,
       toolbarActions,
       interactableToolbar,
-      showNavDrawer,
       cardClasses,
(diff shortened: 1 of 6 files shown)


src/components/charts/shared/navDrawer/index.jsx
```diff
@@ -55,44 +54,38 @@ class NavDrawer extends React.PureComponent {
...
+            desktopDrawerType={NavigationDrawer.DrawerTypes.PERSISTENT_MINI}
+            overlay={false}
+            contentClassName={contentClasses}
+          >
+            {children}
+          </NavigationDrawer>
+        </Card>
 
-        {!showNavDrawer && children}
+        {children}
```
blnkt: Should take this line out altogether.  As is every page with a NavDrawer has duplicated content below the NavDrawer

blnkt (approved):

PR #295, opened 2020-12-10
[F] Icon Override for ChartSwitcher

parkerkaufmann:
EPO-3427


blnkt (changes requested):
Excellent.  I didn't even think about the wrinkle of nested widgetness.  Thank you for accounting for that.  Can we update the styles such that these like "text" icons look more like how the icons look?  So, dark green text over the gray when inactive, white over dark green when active.

PR #296, opened 2021-01-04
[F] New Expanding Universe Images & Page

parkerkaufmann:
Added 3 new images and one new page.
EPO-4012


blnkt (approved):
Excellent.  One lil prob on line 2 of the new page you made: that id needs to be unique but was the same as another page's id cause copy/paste, so the page routing was not working quite right.  But no biggy.  I already went ahead and fixed.  Otherwise good to go: images look  great and placeholder text A-ok.  Ardis and Jose will be so pleased to see this "live."

PR #301, opened 2021-02-11
[EPO-4239] Copy changes

georgegomez144:


blnkt (changes requested):
All those `reduxcache...blah` files are junk.  Is a symptom of getting your local stuck in some sorta state where `gatsby develop` got run twice, or else you ran `gatsby clean` while you had the dev server running, or I don't even know what else.  Sufficed to say we don't want them, and if you ever see something like it delete it, and it def shouldn't get checked in.  

Once you re-commit without those cache files, and you address the acknowledgements issue I noted inline, I'll merge this work and deploy to netlify, and then I'll ask you to present it to Ardis for her to review and ensure we applied her changes where and how she intended.

blnkt (approved):
So you have a new commit that deleted the cache files and made that quick change to the acknowledgments page.  That's fine, but has one unintended side effect.  Because the addition and deletion of those files is still in commits, all those files still get stored in git forever.  These aren't huge files, but this kind of hidden bloat can add up over time, and is really hard to weed out after the fact.  To avoid this pitfall, I rolled back your commits, but left the work intact and unstaged `git reset HEAD~2`, and then deleted the cache files `rm -rf reduxcacheblahs`, and then created a new commit with all of your copy changes, but that doesn't know those bad files even ever existed.  

I'm gonna merge this PR onto master, and then deploy these changes to https://surveyingthesolarsystem.netlify.app/.  After I do, I'll ping you and you should handle communication & collaboration with Ardis.  If after she reviews there are further edits needed, please continue to track against EPO-4239, but create a new branch & PR.  After Ardis indicates our work well reflects her content revisions I'll close the ticket.

PR #304, opened 2021-02-16
[EPO-4024] Fix TOC overflow hidden issue

georgegomez144:
Utilize positioning of the TOC and apply a media query to catch screen size changes.

### DIFF
--- a/src/components/site/tableOfContents/table-of-contents.module.scss
+++ b/src/components/site/tableOfContents/table-of-contents.module.scss
@@ -70,7 +70,6 @@
 }
 
 .table-of-contents {
-  $pageNavHeight: 168px;
   z-index: 14;
   padding-top: calc(#{$siteHeaderHeight} + 10px);
 
@@ -82,10 +81,12 @@
 
   :global {
     .md-list {
-      height: 100%;
-      max-height: calc(100vh - #{$pageNavHeight});
+      height: calc(100vh - 324px);
       padding-top: 0;
-      padding-bottom: $pageNavHeight;
+      padding-bottom: 0;
+      @include respond($break80, max) {
+        height: calc(100vh - 330px);
+      }
     }
   }
 }


src/components/site/tableOfContents/table-of-contents.module.scss
```diff
@@ -70,9 +70,13 @@
...
 
 .table-of-contents {
-  $pageNavHeight: 168px;
   z-index: 14;
   padding-top: calc(#{$siteHeaderHeight} + 10px);
+  position: fixed;
+  bottom: 326px;
+  @include respond($break80, max) {
+    bottom: 330px;
+  }
```
blnkt, line 79: Is already `position: fixed` because of `md-drawer--fixed` class.  But also, I think this approach of "raising" the bottom of this el off the bottom of its parent el is maybe a bit of a hack.  Would you tell me a bit about how/why you landed on this solution?  I'm gonna make a suggestion below that I think is in the spirit of your solution, but coming at it in a maybe more conventional way?

src/components/site/tableOfContents/table-of-contents.module.scss
```diff
@@ -82,10 +86,10 @@
 
   :global {
     .md-list {
-      height: 100%;
-      max-height: calc(100vh - #{$pageNavHeight});
       padding-top: 0;
-      padding-bottom: $pageNavHeight;
+      position: absolute;
+      top: 270px;
+      width: 100%;
```
blnkt, line 92: So above you offset `.table-of-contents` by ~330px in order to get its child, `.md-list`, the el with overflow set to scroll, scroll over the desired vertical distance.  I think that 330px is not an accident, but is in fact the height of the toc + the height of the main header + the height of the page nav.  And so, I suspect, you can accomplish the same visual effect as you have now, just by setting the height of this el explicitly to the height of the screen (100vh) less the over fixed pos els (the height of the toc + the height of the main header + the height of the page nav), and removing the new positionoing stuff youo have on `.table-of-contents`.  So like `height: calc(100vh - 330px)` (326px on wide screens).  Or altogether 

```
.table-of-contents {
   z-index: 14;
}

.md-list {
   height: calc(100vh - 326px);
   padding-top: 0;
   padding-bottom: 0;

   @include respond($break80, max) {
      height: calc(100vh - 330px);
   }
}
```

I think....?

blnkt (changes requested):
Has the desired effect, but I'm proposing a more straight-forward solution.  Please evaluate and test, and let me know if it seems better or worse to ya.

src/components/site/tableOfContents/table-of-contents.module.scss
georgegomez144: After applying your proposed solution

Code:
image


Firefox:
image

Chrome:
image

georgegomez144: That solution works great. Just need to keep this line in `.table-of-contents`
```
.table-of-contents {
  z-index: 14;
  padding-top: calc(#{$siteHeaderHeight} + 10px);  <----- this line
...
```
I also adjusted the height px values a little to account for the teeny-tiny space at the bottom.

src/components/site/tableOfContents/table-of-contents.module.scss
georgegomez144: I wanted to be more 'forceful' with this approach as not to leave any room for misinterpretation from any browsers.

src/components/site/tableOfContents/table-of-contents.module.scss
blnkt: yes.  good catch.

PR #309, opened 2021-02-23
[EPO-4343] Update logo and favicon

georgegomez144:

## What this change does ##

The logo and favicon have now been updated to the current (approved) logo designs by José.

## Notes for reviewers ##

Updates should be seen in the `gatsby-config.js` for the favicon updates and in the `src/layouts/index.jsx` for the logo update.

- 10 = "I think it works, I can explain how, but only over screenshare"
Scale of detailed review would be about a 2.

## Testing ##

Change tested by viewing the webpage header and tab in both dark and light tabs.

## Checklist ##

### Before:
image

### After:
image


blnkt (changes requested):
Logo in the header looks great.  We actually have a dedicated image to use for the favicon.
Rubin favicon 32px

I've added it as an attachment on [the ticket](https://jira.lsstcorp.org/browse/EPO-4343).  Please update to use it.

PR #310, opened 2021-02-24
[EPO-4342] Add connecting lines from MW to each galaxy

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EPO-4342

## What this change does ##

Add connecting lines from the Milky Way galaxy to each individual galaxy. 
In order to create a line for each connection individually, I had to build an array of objects and spread them into the series array.

## Notes for reviewers ##

Updates can be found in the `/components/charts/galaxiesPosition3D/index.jsx`

**Requesting a level 5 review**

## Testing ##

Visually ? :) 

## Checklist ##

### Before:
image

### After:
image

### DIFF
--- a/src/components/charts/galaxiesPosition3D/index.jsx
+++ b/src/components/charts/galaxiesPosition3D/index.jsx
@@ -30,8 +30,27 @@ class GalaxiesPosition3D extends React.PureComponent {
     });
   }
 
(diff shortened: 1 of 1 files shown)


blnkt (approved):
Works like a dream

PR #312, opened 2021-03-01
[EPO-4339] Update #glossary terms to the production site

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EPO-4339

## What this change does ##

Found and replaced all instances of glossary terms established in the investigations in focus (expanding and solar).

## Notes for reviewers ##

Change can be found through a search for #glossary (should yield no results) and a search for "http://rubineducation.org/glossary" as this is the URL I have attached the terms to.

- 10 = "I think it works, I can explain how, but only over screenshare"
- 
Review level 3 requested

## Checklist ##


blnkt (changes requested):
Thank you for tracking down all these terms and linking them to their definitions on the ed site.  There are 2 issues I'd like to see you address before we merge this work in:
1. Glossary links should look the same as other links
2. Should be able to see the glossary term when you land on the ed site page (see screenshot)

For 1. this should be as straightforward as stripping the bold tag off the terms.  I like the way the bold looks, but when these links interspersed amongst regular (not glossary term) links, it's gonna be confusing for the user if they're not all styled like links.

For 2.  I think the issue is that the scroll to behavior on your #hashtag style links is not properly compensating for some vertical height eaten up but the header and such.  It's not an issue if the page is already loaded, but if you click through [one of your links](http://rubineducation.org/glossary#hubble_plot) in the investigation, it's loading a new page and it's not landing on the term as expected.  I think you have two options.  You can either revisit your work in the Ed Site development and figure out how to affect the expected behavior, or you can use a different style of link to link to glossary terms on their own dedicated pages (a newly added feature I introduced to get previews and the glossary links in the Teacher Guides working correctly).  These "dedicated" links are shaped similarly to what you have: `http://rubineducation.org/glossary/hubble-plot` instead of `http://rubineducation.org/glossary#hubble_plot`.  I think I would suggest using the "dedicated" links strat since it's probably gonna be a faster fix but I'm interested in your opinion.  Ping me in Slack with what you're thinking and we'll go from there.

https://user-images.githubusercontent.com/8166060/109549709-2e3d3600-7a83-11eb-8d78-cc501ce59d9d.mov

georgegomez144:
1. For the bolding of the links, i was following Ardis' suggestion of having them like that. I can certainly change them to not have them bold any longer.
2. That bug is very interesting. I'm only seeing it in FireFox and Opera. The -webkit- browsers (chrome, safari, etc..) all work as expected. How would you like to move forward with this? Should I try and find an non-Foundation solution? or?

blnkt:
1.  Yes, please unbold these links.  As I said, when these links are interspersed amongst regular (not glossary term) links, it's gonna be too confusing for the user if they're not all styled the same and all look like links (i.e. if some links look like links, and some links look like normal bold text, it's likely the user will only ever click on the links that look like links.

2.  This issue presents itself in every browser I've tried.  As I said, I think you have two options. You can either revisit your work in the Ed Site development and figure out how to affect the expected behavior, or you can use a different kind of link to link to glossary terms on their own dedicated pages.  I suggested the latter is probably a faster/easier solution, but if you can suggest compelling reasons to fix on the ed site side I'm interested in your perspective.

blnkt (changes requested):
1.  Yes, please unbold these links.  As I said, when these links are interspersed amongst regular (not glossary term) links, it's gonna be too confusing for the user if they're not all styled the same and all look like links (i.e. if some links look like links, and some links look like normal bold text, it's likely the user will only ever click on the links that look like links.

2.  This issue presents itself in every browser I've tried.  As I said, I think you have two options. You can either revisit your work in the Ed Site development and figure out how to affect the expected behavior, or you can use a different kind of link to link to glossary terms on their own dedicated pages.  I suggested the latter is probably a faster/easier solution, but if you can suggest compelling reasons to fix on the ed site side I'm interested in your perspective.

blnkt (approved):
Only thing still missing was my suggested fix for the glossary term urls.  I went ahead and did a find/replace and updated them to all link to the dedicated glossary term pages on the ed site.

PR #315, opened 2021-03-04
[EPO-4342-fix] Connective lines in GalaxiesPosition3D always have correct origin

georgegomez144:

## What this change does ##

This is an update to how the data is loaded for the "new" line graph under the plotted points. Originally, the lines were being built off of the `noLabels` array when it should have been build off of the raw `data` array initially provided. This allows all lines to be created correctly. 
## Notes for reviewers ##

A "catch" for the currently selected galaxy to be the point of origin was implemented in the function `createLineData` that listens for the length of the `noLabels` array in order to provide the correct point of origin.

**Review level 3 requested**

## Testing ##

This change can be tested by clicking through the menu options on the left chart to switch the point of origin.

## Checklist ##


blnkt:
Your description in "What this change does" doesn't really reflect what change you made, and also I encourage you to include that section with a statement that captures the outcome of the change rather than your process.  I don't feel like your first sentence, "This is an update to how the data is loaded for the "new" line graph under the plotted points" makes sense, even in context.  And the last sentence, "This allows all lines to be created correctly" does not tell the reviewer what "correct" means.  I think a clearer more comprehensive "What this change does" might be something like:

> Fixes a bug where the galaxy used as the origin for the connecting lines in GalaxiesPosition3D was not updating when the user selected a new galaxy.  The problem was `getConnectingLines` was being passed the `noLabels` array when it should have receives the `data` array.  Passing `getConnectingLines` the correct array resulted in connecting lines drawn between the origin galaxy and each other galaxy regardless of which galaxy the user selected.

PR #316, opened 2021-03-04
[EPO-4419] Fix lazyloading

blnkt:
JIRA: [EPO-4419](https://jira.lsstcorp.org/browse/EPO-4419)
## What this change does ##

Re-implements timeout on lazy-loaded images.  The timeout should ensure that the image becomes visible after 4 secs, even if the `onload` has not fired, even if the image has not finished loading.

## Notes for reviewers ##

This is a change we'll have to keep an eye on on staging due to challenges testing.

1 = "I barely need review on this"

## Testing ##

Tested locally by throttling network in chrome devTools, and on staging by checking when `console.log`s fired in the timeout callback vs the `onload` handler.  Console.logs have since been removed, but can still test on staging by observing when or if images in the investigation load in 4 secs or less.

## Checklist ##

- [X] This change impacts several investigations (e.g. the change affects reused styles, widgets, pages, QAs, utility methods, etc.)

### DIFF
--- a/src/components/site/imageLoader/index.jsx
+++ b/src/components/site/imageLoader/index.jsx
@@ -15,21 +15,16 @@ class ImageLoader extends React.PureComponent {
       loaded: false,
     };
 
-    // this.timeFallback = null;
-    // this.setTimeoutFallback();
+    this.timeFallback = setTimeout(this.handleLoading, 4000);
   }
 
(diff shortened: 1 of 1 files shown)


georgegomez144 (commented):
Used chrome's throttle method to test and works as described.

PR #317, opened 2021-03-04
[EPO-4351] Add QA Review page

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EPO-4351

## What this change does ##

This new page allows for the teachers and students to review all answers within the investigation. This will also allow for the eventual printability of this page.

## Notes for reviewers ##

There are many places that I've touched within this branch. I've added a few new classes within the `questions/styles.module.scss` that apply some styling that the question labels were creating. I had to do this in order to keep the flow of the investigations as is. This required me to dive into each question component and adjust the code in order to render the proper html element on the QA Review page. I've also made sure that this change did not affect the way the questions were rendering and functioning in the investigation.

Also, a heads up, I had to use `!important` within the css to force some styling updates. Can you take a look at this to be sure this is the only way. From my angle, I believe it is.

**I'm going to request a review level of 6 to ensure that all bit touched are correct and implemented as expected.**

## Testing ##

If I remember correctly, there are calculator tests created, but none other than those.

## Checklist ##

### After:

https://user-images.githubusercontent.com/8799443/110026666-29c88580-7cee-11eb-80da-fc711744ac2d.mp4

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -253,6 +253,23 @@ exports.createPages = async ({ graphql, actions }) => {
     investigations.forEach(investigation => {
       const { id } = investigation;
       createPage({
+        path: `/${id}/qa-review/`,
+        component: path.resolve(`./src/containers/QAReviewContainer.jsx`),
+        context: { investigations, investigation: id, env: INVESTIGATION },
+      });
+    });
+  } else {
+    createPage({
+      path: '/qa-review/',
+      component: path.resolve(`./src/containers/QAReviewContainer.jsx`),
+      context: { investigations, env: INVESTIGATION },
+    });
+  }
+
+  if (isAll) {
+    investigations.forEach(investigation => {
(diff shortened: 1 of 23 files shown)


blnkt (changes requested):
FYI I pushed new stuff to this branch so I think you'll need to reset your local to match this remote.

- CompoundInput is not rendering correctly:
Screen Shot 2021-03-08 at 10 22 21 AM
- Prompt styling doesn't match rest of questions' treatment: 
Screen Shot 2021-03-08 at 11 07 30 AM

blnkt (changes requested):
Looks good on the review page, but in the investigation proper the prompt layout is busted because you wrapped your label in a p tag.  
Screen Shot 2021-03-09 at 11 34 57 AM

Please fix.

blnkt (approved):

PR #320, opened 2021-03-08
[EPO-4405] Add tables to new GalaxySupernova pages

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EPO-4405

## What this change does ##

This change adds tables to the new pages created from [EPO-4404](https://jira.lsstcorp.org/browse/EPO-4404) that stores and persists the `distance` and `velocity` values from the selection of the new GalaxySupernova widget on each page.

## Notes for reviewers ##

Take a look at how the data is being returned from within the widget as well as how the `WithQAing.jsx` component is returning the content for the question component.

Requesting a review level of **7**

## Testing ##

This update can be tested by selecting a galaxy and supernova on `interpreting-a-hubble-plot-1/` then clearing out the selection with the "clear" option on the question. After the selection is made, moving on to the new page to see your selection persist in the table. Make sure that each page in the series works as expected.

## Checklist ##

- [x] This is a breaking change: changes in page data (`/src/data/pages/PAGE.json`) or scientific data (anything in `/static/`) and I will notify Product Marketing when it is in prod and user-visible.
- [x] This change impacts several investigations (e.g. the change affects reused styles, widgets, pages, QAs, utility methods, etc.)

## Screenshots (optional) ##
image

### DIFF
--- a/src/components/widgets/widgets-utilities.js
+++ b/src/components/widgets/widgets-utilities.js
@@ -1,6 +1,7 @@
 import SupernovaSelectorWithLightCurve from '../../containers/SupernovaSelectorWithLightCurveContainer.jsx';
 import GalaxyScrambler from '../../containers/GalaxyScramblerContainer.jsx';
 import GalaxySelector from '../../containers/GalaxySelectorContainer.jsx';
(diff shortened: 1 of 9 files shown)


blnkt (approved):
Is workies.

PR #322, opened 2021-03-10
[EPO-4406] Create the new GalaxyHubblePlotContainer.jsx

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EPO-4406

## What this change does ##

This change transforms the GalaxySelectorContainer into the GalaxyHubblePlotContainer by stripping out and updating core pieces to handle having only the HubblePlot component with its interactions. This container will still handle having multiple galaxies in order to plot the corresponding points.

## Notes for reviewers ##

Changes should be found in the new GalaxyHubblePlotContainer.jsx as well as an update to the galaxySelectorUtilities.js. The page this new widget lives on is `interpreting-a-hubble-plot-2/`. On this page, you should see a table with your previously populated distance and velocity values from the previous pages. This will be the information in which will define your interaction with the new widget.

Review level requested is **5**

## Testing ##

The change can be tested by either selecting the galaxy you wish to plot using the sidebar navigation or the Previous and Next buttons, then plotting the values.

## Checklist ##

## Screenshots (optional) ##
image
NewGalaxyHubblePlotContainer (1)


src/components/charts/galaxySelector/galaxy-selector.module.scss
```diff
@@ -145,3 +145,7 @@ nav {
     }
   }
 }
+
+.hubble-plot-container {
+  padding: 20px 20px 20px 80px;
```
blnkt, line 150: please use the `$minPadding` variable instead of `20px`

blnkt (changes requested):
- Please update the "next" and "rev" galaxy button styles so text doesn't overflow out on screens as small as 1024px width: 
Screen Shot 2021-03-15 at 11 09 12 AM
I think should be as straight-forward as adding height: 100%, but be sure to check across browsers.

- I think this widget should still have an associated accordion type question with saved answer.  Clicking that question's clear button should reset the hubble plot, just as the previous iteration of this widget did, and just as the trendline does on subsequent pages
- Remove unused methods from GalaxyHubblePlotContainer.  I think this is just `onBlinkChange`.
- Remove the console.log I left in there (whoops)
- How the data array in the initial response is being mutated feels wrong in light of no longer needing alerts or images....I think.  Take a look at see if this component has more info in the data than it is actually making use of.
- I added two commits: 1 tweaks the padding on the y axis label, the other sets activeGalaxy when user selects point.  Please verify both these features are working as expected.

blnkt (changes requested):
I'm not sure, did you mean to request my review of this PR @georgegomez144?  I see one new commit addressing one of my comments, but there are several others left unaddressed.  Intentional?  Missed my previous comment?  Making some other plan?

georgegomez144:
Sorry, fixing those now.

georgegomez144:
@blnkt struggling to get this accordion question type to load correctly for the new hubble plot. I'm guessing it has something to do with accessor???

blnkt:
> @blnkt struggling to get this accordion question type to load correctly for the new hubble plot. I'm guessing it has something to do with `answerAccessor`???

@georgegomez144 So your work as is is currently already propagating an answer for a question with the ID 252.  This answer is created in the `userHubblePlotCallback` and dispatched by `updateAnswer` in `WithQAing.jsx`.  You can console.log the answer there to see what it looks like.  Looking at the `WithQAing` file it looks like there’s already a `hubblePlot` `answerAccessor`.  I don’t think it’s currently used…looks like I set it up for this purpose and then it got supplanted by the galaxy `answerAccessor` at some point…in any case, use that one or make a new one, make a question on the page with the corresponding `id` & `answerAccessor`, and then you should start to see something?   Hopefully that gives you enough to go on?  Let me know if you're still running into any blocker.

PR #325, opened 2021-03-11
[EPO-4402] Disable TOC links if earliest questions not answered

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EPO-4402

## What this change does ##

Disable TOC links when the earlier questions have not been answered to keep from "jumping around" in the investigation.
## Notes for reviewers ##

Create a new css class that disables `pointer-events`. This works in tandem with the `disabled` param on the `links` object created in the TOC. I'm also mutated (slightly) the `checkQAProgress` function in the `tableOfContents/index.jsx` that now takes a specific `state` (e.g. 1 or 0) to determine the state of the page in question. I then use this to determine in disabled state of each link per its page. Works nicely with the progress states.

Include an indication of how detailed a review you want on a 1-10 scale. **5 = "Please make sure there are no obvious errors and that you believe it does what it says it does"**

## Testing ##

Testing can be done by trying to navigate through the app without answering the questions and with answering the questions.

## Checklist ##

### DIFF
--- a/src/components/charts/colorMixingTool/index.jsx
+++ b/src/components/charts/colorMixingTool/index.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import isArray from 'lodash/isArray';
+import filter from 'lodash/filter';
 import SliderCustom from '../../site/slider/index.jsx';
 import Select from '../../site/selectField/index.jsx';
 import Button from '../../site/button/index.js';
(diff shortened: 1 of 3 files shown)


blnkt (changes requested):
- Link in TOC for current page (the page the user is looking at) should never be disabled
- Seems like there are some edge cases related to Prompt and Accordion Qs.  In the case of the prompts it seems like they are always skippable.  Not a problem if the page has other questions on it, but the coloring investigation has several pages that only have a prompt.  In the case of accordions I found it was inconsistently accounting for the question being considered answered or not.  I'm not going to recommend trying to address the accordion issue at this time, but in the case of the prompts it would be super nice if we could have those produce answers (even if the answer isn't visible by the user) so they can be tabulated in the same manner as the other kinds of questions.
- I added a commit that refines the checkProgress method.  You had changed it to take the state param.  As far as I can tell it could be accomplished with the old version of that method, setting its output to a variable `const allQsComplete = this.checkQAProgress(id)` and then using `allQsComplete` or `!allQsComplete` where needed.  But perhaps I'm missing some nuance.  Please evaluate.

PR #327, opened 2021-03-11
[EPO-4394] Add checkpoint graphic to investigation

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EPO-4394

## What this change does ##

Creates the `Checkpoint` component that uses the layout object in the json to place the widget on the page.

## Notes for reviewers ##

Can find this change in a few many places. This touches quite a bit of the base code all the way down to the fragments. I've included this update in where the page blocks exist.

## Testing ##

This change can be tested by adding a checkpoint or checkpoints and changing the values in the layout to move it around the page.

## Checklist ##

## Screenshots (optional) ##
image


blnkt (changes requested):
Looks great.  Works great.  Good anticipating of treating this as a block, and the flexibility therein.  Only thing I'll call out is that you have two Checkpoint svgs floating around.  Not sure which one is "the one" put please remove the one that's not.

georgegomez144:
Remove the non-needed checkpoint.svg 👍

PR #328, opened 2021-03-15
[EPO-4399] Add tables to QA Review page

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EPO-4399

## What this change does ##

This commit updates the GraphQL query to use the PageContainer's query in order to set up and pull in all elements of the investigation. This also includes the `ObservationsTable` component in order to render the "now available" tables in the QA Review page.

## Notes for reviewers ##

Changes can be found on the `qa-review` page. Look for the tables loaded at the bottom of the QA Review page.

Include an indication of how detailed a review you want on a 1-10 scale. **5 = "Please make sure there are no obvious errors and that you believe it does what it says it does"**

## Testing ##

Testing this change can be seen by "filling in" the tables through the investigation and checking to see the answer is correctly populated in the QA Review page.

## Checklist ##

### Before:
image

### After:
image

### DIFF
--- a/src/components/qas/styles.module.scss
+++ b/src/components/qas/styles.module.scss
@@ -1,3 +1,4 @@
+/* stylelint-disable declaration-no-important */
 .qa {
   + .qa {
     margin-top: $minPadding / 4;
@@ -28,14 +29,14 @@
   .qa-review-label-post,
   .calculator-label,
   p {
-    font-size: 16px !important; /* stylelint-disable-line declaration-no-important *//* stylelint-disable-line declaration-no-important *//* stylelint-disable-line declaration-no-important */
-    font-weight: $medium !important; /* stylelint-disable-line declaration-no-important *//* stylelint-disable-line declaration-no-important */
-    color: $basePrimary !important; /* stylelint-disable-line declaration-no-important */
+    font-size: 16px !important;
+    font-weight: $medium !important;
+    color: $basePrimary !important;
   }
 }
 
 .qa-review-block {
-  margin: 0 !important; /* stylelint-disable-line declaration-no-important */
+  margin: 0 !important;
   font-size: 16px;
   font-weight: $medium;
   color: $basePrimary;
@@ -59,7 +60,7 @@
   .qa-review-inline-input {
     display: inline-block;
     max-width: 180px;
-    margin-top: 0 !important; /* stylelint-disable-line declaration-no-important */
+    margin-top: 0 !important;
     text-decoration: underline;
   }
 
@@ -74,4 +75,9 @@
       text-decoration: underline;
     }
   }
+}
+
+.qa-review-table-container {
+  max-width: 800px;
+  margin: 2 * $minPadding $minPadding;
 }
--- a/package.json
+++ b/package.json
@@ -80,7 +80,8 @@
     "serve": "gatsby serve",
     "test": "jest --verbose",
     "watch-test": "jest --watch --verbose",
-    "clean": "gatsby clean"
+    "clean": "gatsby clean",
+    "develop": "yarn clean && yarn start"
   },
   "repository": {
     "type": "git",
--- a/src/containers/QAReviewContainer.jsx
+++ b/src/containers/QAReviewContainer.jsx
@@ -6,8 +6,10 @@ import { graphql, Link } from 'gatsby';
 import find from 'lodash/find';
 import filter from 'lodash/filter';
 import flattenDeep from 'lodash/flattenDeep';
+import ObservationsTable from '../components/charts/shared/observationsTables/ObservationsTable';
 import QAs from '../components/qas';
 import Button from '../components/site/button/index.js';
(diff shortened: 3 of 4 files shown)


src/components/qas/styles.module.scss
```diff
@@ -28,8 +28,8 @@
   .qa-review-label-post,
   .calculator-label,
   p {
-    font-size: 16px !important; /* stylelint-disable-line declaration-no-important *//* stylelint-disable-line declaration-no-important *//* stylelint-disable-line declaration-no-important */
-    font-weight: $medium !important; /* stylelint-disable-line declaration-no-important *//* stylelint-disable-line declaration-no-important */
+    font-size: 16px !important; /* stylelint-disable-line declaration-no-important */ /* stylelint-disable-line declaration-no-important */ /* stylelint-disable-line declaration-no-important */
+    font-weight: $medium !important; /* stylelint-disable-line declaration-no-important */ /* stylelint-disable-line declaration-no-important */
```
blnkt, line 32: Not sure what's going on here but got too many disable linter rule comments goin on here.  Maybe it doesn't like neighboring lines having the disable-line style comment?  Try the other style of comment where you tell it when to start and stop ignoring the rule?  [More info here](https://stylelint.io/user-guide/ignore-code).

src/components/qas/styles.module.scss
```diff
@@ -75,3 +75,8 @@
     }
   }
 }
+
+.qa-review-table-container {
+  max-width: 800px;
+  margin: 40px 20px;
```
blnkt, line 81: Please use the $minPadding variable instead of hardcoding 20px or 2 * 20px.

blnkt (changes requested):
Just a couple quick changes, otherwise works as advertised.  I will note, you've put yourself in a position where you're going to be redoing your business logic here once you address [EPO-4350](https://jira.lsstcorp.org/browse/EPO-4350).  Sorry if we didn't discuss in what order these associated tasks ought be addressed.  Don't think it should slow you down too much, just something to keep in mind the next time you are breaking a larger deliverable into smaller steps and deciding in what order to address those steps.

georgegomez144:
Completely in the dark about how I'm going to get all of the questions in order with the tables and widgets in the right place...totally lost on that right now...

blnkt:
> Completely in the dark about how I'm going to get all of the questions in order with the tables and widgets in the right place...totally lost on that right now...

You're grabbing all the pages already.  Each page has an `order` value that we introduced just for the purpose of sorting them in the order in which they appear in the investigation.  So if you sort your array of pages on that value they'll be in order.  Then you loop through your sorted array and have nested loops for each type of page content and you're done.  I think we decided the questions always come first, then the tables, then the widgets, but I may be misremembering.

src/components/qas/index.jsx
```diff
@@ -17,11 +17,11 @@ class QAs extends React.PureComponent {
 
     return (
       <div className="qas">
-        {questions.map(question => {
+        {questions.map((question, qid) => {
           const { question: q, number } = question;
           const primeQ = q[0];
           const { id, questionType } = primeQ;
-          const key = `qa-${id}`;
+          const key = `qa-${id}-${qid}`;
```
blnkt, line 24: I didn't notice this last time around but if you're doing this because you're getting an error about unique keys, that's a good error and you should change whatever the duplicate id is instead of adding the index onto the key to make it unique. In other words, it's important for keys to be unique, but for our answers to appropriately line up with our questions, it's also important for the question ids to be unique. Please update.

blnkt (changes requested):
one more quick change and otherwise this looks good to go.

src/components/qas/index.jsx
georgegomez144: Do you mean to say, update _all the questions_ in the jsons?

georgegomez144: In this specific case, the Coloring the Universe investigation has a widget and question that gets dragged across multiple pages that needs to persist its answer. How should we handle this, then?

blnkt: shouldn't be a problem since keys' uniqueness is scoped to their loop, not the page.

georgegomez144: Interesting, but if that was the case why does the app throw the error?

blnkt (approved):
Thanks for making those changes.  Looks like it's working as expected.  I'm just going to note that the order & sort stuff you did here is not in the scope of this ticket, you have [a whole separate ticket you made to address that aspect](https://jira.lsstcorp.org/browse/EPO-4350).  I can understand wanting to introduce that change as soon as you know about it but 2 things: 
1. now the ticket & PR descriptions don't align with the work you delivered, which could be potentially confusing
2. I don't think the sort func you've written is going to sort stuff how you expect since the `order` values are strings not numbers (e.g. I'm fairly certain some of them look like 12a, 12-1, and the like) and I want to provide feedback on that, but the ordering question numbering stuff belongs to a different ticket/PR, so if I get into it here, then what is that PR for, and maybe you've already started that one, and see point 1

PR #329, opened 2021-03-16
[EPO-4397] Add widgets to the QA Review page

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EPO-4397

## What this change does ##

This change adds the widgets to the QA Review page. Used the same method of the tables to load these widgets. Created a container class to make sure the widget fits in the bounds of the page.

## Notes for reviewers ##

Well, the QA Review page is the place to be when reviewing this PR.

There is a style change to the PrismTool that was breaking on the QA Review page, and well, this change makes the layout in the investigation a little cleaner.

**5 = "Please make sure there are no obvious errors and that you believe it does what it says it does"**

## Testing ##

n/a

## Checklist ##

### DIFF
--- a/src/components/charts/prismWidget/prism-widget.module.scss
+++ b/src/components/charts/prismWidget/prism-widget.module.scss
@@ -1,4 +1,5 @@
 .content-wrapper {
+  position: relative;
   margin: auto;
   overflow: hidden;
   text-align: center;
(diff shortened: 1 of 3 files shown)


blnkt (approved):
I did not test with every kind of available widget, but the approach seems sound, and works with the ones I did test it with in the Expanding Universe Investigation.

PR #331, opened 2021-03-17
[EPO-4350] Order and add number to pages in QA Review

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EPO-4350

## What this change does ##

This change ensures that questions/pages load in order as shown in the investigation on the QA Review page. This change also adds the question numbers. I used a similar method found in the `PageContainer.jsx` file.

## Notes for reviewers ##

There were a few `.json` pages that were updated (the `order` field) to make sure that the pages/questions are loaded correctly in order.

Include an indication of how detailed a review you want on a 1-10 scale: **5 = "Please make sure there are no obvious errors and that you believe it does what it says it does"**

## Testing ##

n/a

## Checklist ##

### DIFF
--- a/src/components/qas/styles.module.scss
+++ b/src/components/qas/styles.module.scss
@@ -43,6 +43,7 @@
 
   .answer-label,
   .answer-label p {
+    display: inline-block;
     @extend .qa-review-block;
   }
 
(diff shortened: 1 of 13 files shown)


blnkt (approved):
Good good stuff.

PR #333, opened 2021-03-18
[EPO-4468]  HubblePlotter points are draggable

blnkt:
JIRA: [EPO-4468](https://jira.lsstcorp.org/browse/4468)

## What this change does ##
- If active galaxy has no point placed user should see "ghost point" and be able to place point on 
- If active galaxy has a point already user should be able to adjust its placement by dragging and dropping
- If user clicks or drags a non-active galaxy point that point and its corresponding galaxy should become active
- This new interactivity should be scoped to the HubblePlotter only (not the normal HubblePlot or the HubblePlotWIthTrendline or the combo with the 3D stufff)

## Notes for reviewers ##
To work around the drag event colliding with the normal click event I had to get creative in terms of what a drag-start and drag-end events trigger, especially in the absence of any drag fired in between.  Feels...a bit frail, but held up in testing.

This is a 7.

## Testing ##
I tested in the 4 contexts I could think of: Hubble/Plotter, hubblePlot, HubblePlotWIthTrendline, and the combo of it and the 3D positional graph.  But I tested most exhaustively in the former.  Please check across all these contexts and any I may be missing in the expanding universe investigation.  Please test all available interactivity and UI (point placement, side menu, full sidebar menu, and next/previous buttons, etc) in the widget when there are no points and/or no answer (fresh start), when some but not all points have been placed, when all points have been placed, after clearing points through QA UI.  Please test across browsers.
There does seem to be some edge cases where the ghost point disappears during drag.  Please provide indication of reproducibility and and thoughts on fix.

## Checklist ##

- [ X ] This change is isolated to a specific page or Component (i.e. it's a one-off)

### Before:
https://user-images.githubusercontent.com/8166060/111666671-2abee400-87d1-11eb-9bda-9ff17a36c209.mov

### After:
https://user-images.githubusercontent.com/8166060/111671625-5ee8d380-87d6-11eb-9387-92ac1169da01.mov

### DIFF
--- a/src/assets/stylesheets/components/charts/_tooltip.scss
+++ b/src/assets/stylesheets/components/charts/_tooltip.scss
@@ -3,7 +3,7 @@
   padding: 8px;
   font: 12px sans-serif;
   pointer-events: none;
-  background: lightsteelblue;
+  background: $lightBlue;
   border: 0;
   border-radius: 8px;
 }
--- a/src/components/charts/hubblePlot/Points.jsx
+++ b/src/components/charts/hubblePlot/Points.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import includes from 'lodash/includes';
+import find from 'lodash/find';
 import classnames from 'classnames';
 import Point from './Point.jsx';
 import { notActive, invisible } from './hubble-plot.module.scss';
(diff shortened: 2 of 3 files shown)


georgegomez144 (approved):
Works as advertised!!! Very smooth, I might add.

PR #341, opened 2021-04-05
[EPO-4471] Update data for OrbitalViewer

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EOP-4771

## What this change does ##

This change adds the json data, provided by Lauren, to have a consistent naming convention across all OrbitalViewer objects.

## Notes for reviewers ##

This updates the index used for the name parameter in the object details.

Include an indication of how detailed a review you want on a 1-10 scale:   5 = "Please make sure there are no obvious errors and that you believe it does what it says it does"

## Testing ##

n/a

## Checklist ##

- [a] This is a breaking change: changes in page data (`/src/data/pages/PAGE.json`) or scientific data (anything in `/static/`) and I will notify Product Marketing when it is in prod and user visible.
- [a] This change is isolated to a specific page or Component (i.e. it's a one-off)

### Before:
image

### After:
image


src/components/visualizations/orbitalViewer/OrbitalDetails.jsx
```diff
@@ -13,7 +13,8 @@ import {
 import Button from '../../site/button/index.js';
 
 function OrbitalDetails({ data, velocity, type }) {
-  const { H, a, i, e, Principal_desig: name, Earth_moid: moid } = data || {};
+  const { H, a, i, e, name, Principal_design: pdName, Earth_moid: moid } =
```
blnkt, line 16: Typo: `Principal_design` should be `Principal_desig`.  As is is breaking the name in details pane toward the end of the investigation: pages `classifying-objects`, `classifying-objects-1`, and `classifying-objects-2`.  Alternatively if all the new data uses `name` instead of `Principal_desig` could update the data for the aforementioned pages to use `name` as well.

blnkt (changes requested):
Nearly there, just a typo and whatnot.

PR #342, opened 2021-04-05
[EPO-4549] Add new datasets from Lauren and update widget

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EPO-4549

## What this change does ##

This change adds the json data, provided by Lauren, to have a "pre-binned" data option.

## Notes for reviewers ##

This solution implements a duplicate histogram component, `histogramFlat.jsx` in order to introduce a "flat" styled "pre-binned" data set.

## Testing ##

n/a

## Checklist ##

## Screenshots (optional) ##
n/a


blnkt (changes requested):
The tooltips no longer show the ranges of values for each bar when hovered/selected, please correct. 
Screen Shot 2021-04-05 at 9 11 01 PM

Otherwise seems to be working as expected.  After fixing the above please rename `histogramFlat` directory to `histogram` and delete the superfluous (old) directory of the same name.  For future reference, when duplicating a directory to work on a refactor like this it's usually more convenient to edit the original and treat the duplicate as the "old" version, that way git can track the changes.  When the changes are made to a new file git treats the whole file as new changes and and such it's hard to review (i.e. it's hard to identify what changes were made).  Sorry if I misspoke when discussing this strategy earlier today.

PR #354, opened 2021-04-22
[EPO-4651] Update Color Filter Tool to use the canvas element

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EPO-4651

## What this change does ##

No longer will it use the 'background-image' property (since that does not render in printer view). Now comes the time for the "CANVAS" element, which does render as asked.

## Notes for reviewers ##

Still a bit buggy since, the filtered images won't load "right-off-the-bat" when previously answered.

## Testing ##

n/a

## Checklist ##

### DIFF
--- a/gatsby-browser.js
+++ b/gatsby-browser.js
@@ -1,11 +1,11 @@
-import React from 'react';
-import ReactDOM from 'react-dom';
-import axe from 'react-axe';
+// import React from 'react';
+// import ReactDOM from 'react-dom';
+// import axe from 'react-axe';
 
 import './src/assets/stylesheets/styles.scss';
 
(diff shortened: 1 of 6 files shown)


georgegomez144:
These changes make perfect sense! I just needed to break out the order of how elements were being loaded. I've also updated how this lays out on the QA-Review page and how it lays out in print. I've attached photos below:
**On Local (in Chrome) Canvas Element**
image

**On "staging": Background Image solution**
image

**Print layout in Chrome (local)**
image

**Print layout in Firefox (local)**
image

blnkt (commented):
👍

PR #355, opened 2021-04-27
[EPO-4648] Add "All Questions Answered" notification

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EPO-4648

## What this change does ##

This change adds a new all questions answered alert that notifies the student that all questions on the page have been answered and that they may continue onto the next page. This alert includes a close button and a 'go to next page' button.

## Notes for reviewers ##

Had to add a catch for this notification in the PageNav/index.jsx so that the modal could close. Also, I was able to add the 'go to next page' button into the modal by passing in the next page info into the new component.

## Testing ##

Answer all questions on the page and watch the magic!

## Checklist ##

### After:
image

### DIFF
--- a/src/components/pageNav/index.jsx
+++ b/src/components/pageNav/index.jsx
@@ -8,6 +8,7 @@ import ButtonIcon from '../site/button/ButtonIcon';
 import ArrowLeft from '../site/icons/ArrowLeft';
 import ArrowRight from '../site/icons/ArrowRight';
 import AnswerRequiredAlert from '../site/answerRequiredAlert';
+import AnswersCompletedAlert from '../site/answersCompletedAlert';
 import styles from './page-nav.module.scss';
 
 class PageNav extends React.PureComponent {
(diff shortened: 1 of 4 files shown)


src/components/pageNav/index.jsx
```diff
@@ -99,14 +143,22 @@ class PageNav extends React.PureComponent {
...
         <AnswerRequiredAlert
           showAlert={showAlert && !disableButton ? false : showAlert}
           handleClose={this.handleHideAlert}
         />
+        <AnswersCompletedAlert
+          showAlert={
+            showCompletedAlert && disableButton ? false : showCompletedAlert
+          }
+          baseUrl={baseUrl}
+          nextUrlObj={next}
```
blnkt, line 159: see previous comment about producing link url

src/components/pageNav/index.jsx
```diff
@@ -33,13 +55,35 @@ class PageNav extends React.PureComponent {
...
+    const { allAnswered } = this.state;
+
+    if (allAnswered) {
+      this.setState(prevState => ({
+        ...prevState,
+        showCompletedAlert: false,
+      }));
+    }
+  };
+
```
blnkt, line 86: Should closing the notification depend on the `allAnswered` state?  Is this method ever called when `allAnswered` could be `false` (i.e. wouldn't the notification only be visible if `allAnswered` is `true`)?

src/components/pageNav/index.jsx
```diff
@@ -33,13 +55,35 @@ class PageNav extends React.PureComponent {
     }));
   };
 
+  handleShowCompletedAlert = () => {
```
blnkt, line 58: the "Completed" language prefixing all of the variable and method names related to this alert makes sense to me.  For further clarity please update the other alert's variable and method names with something equally as distinct.  As is it reads now it's like there is some generic all purpose alert getting toggled, and then this new highly specialize alert doing its own thing, which isn't really true to what's goin on here.

blnkt (changes requested):
The following is what design I expected Jose to deliver: Success-answering-v1
Obviously, for the most part, you nailed it even without this design provided.  Good on you. 
 Only real diff is let's have the text read "You answered all of the questions on this page.  Go to the next page when you're ready."  where "Go to the next page"  should be a normal link (not a button on its own line).  Again, sorry for assuming what assets you were receiving instead of double-checking them when they were sent to you.  Just like at the airport: "See Something?  Say Something".  In the future, if an asset or data deliverable or whatever really doesn't seem to align with your own expectations, or expectations I set, please don't hesitate to reach out to me to confirm what's up.

Sorry I didn't notice this when you implemented the disabledButton in the pageNav work you completed previously, but it would appear that compound question type's "fully-answered-ness" is not being correctly tabulated:
Screen Shot 2021-04-27 at 9 11 34 AM
It looks like, in the case of compound questions, the "fully-answered-ness" only depends on populating the first field in the compound question.  I do not know if this is true for both compound selects and compound fill in the blanks.  Curiously, it does not see to affect the the `activeQuestion` state, which is something that is also aware of a question's answered-ness.  Because of this, I would suspect the fault is either in the `shouldDisable` method.  All that said, this problem is not directly related to this notification work (this notification work just happens to be dependent on the validity of the `disableButton` value).  So please create a new ticket to address this fault, and I will merge and close this PR after the other aforementioned issues (and issues called out in inline comments) are addressed.  Thank you much.

src/components/pageNav/index.jsx
```diff
@@ -16,13 +17,34 @@ class PageNav extends React.PureComponent {
...
+  }
+
+  componentDidUpdate() {
+    const { disableButton } = this.props;
+
+    if (disableButton) {
+      this.updateAllAnswersBool(false);
+    } else {
+      this.updateAllAnswersBool(true);
+      this.handleShowCompletedAlert();
```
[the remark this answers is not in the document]
georgegomez144: Seeing as this work would all fit into the new ticket [EPO-4668](https://jira.lsstcorp.org/browse/EPO-4668) I'll make sure this logic get cleaned up in that ticket's work as well.

src/components/pageNav/index.jsx
georgegomez144: This was a strange bug that I came across. I had to implement this in order to ensure that the render method didn't toggle to `false` immediately after toggling to `true`. I was console logging the value of `showCompletedAlert` and it showed to logs (from within the `render()` method) when trying to trigger the alert. Both fired one after the other. The first updated `showCompletedAlert` to `true` and the second reverted back to `false`. I couldn't think of any other way to handle this.

src/components/pageNav/index.jsx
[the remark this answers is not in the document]
blnkt: very well.

blnkt (approved):

PR #360, opened 2021-05-03
[EPO-4668] `shouldDisable` now checks for compoundQuestion type

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EPO-4668

## What this change does ##

This change now checks for when then compoundQuestion type is present. Then the shouldDisable method checks all questions and their answers and returns true or false depending on that evaluation.

## Notes for reviewers ##

n/a

## Testing ##

Go to any compound question type and test all answers. Good page to test is /what-should-my-image-look-like/ in coloring the universe investigation.

## Checklist ##

### DIFF
--- a/src/components/pageNav/index.jsx
+++ b/src/components/pageNav/index.jsx
@@ -18,30 +18,16 @@ class PageNav extends React.PureComponent {
     this.state = {
       showAllRequiredAlert: false,
       showCompletedAlert: false,
-      allAnswered: false,
     };
   }
 
   componentDidUpdate(prevProps) {
-    const { disableButton } = this.props;
-    const { disableButton: prevDisableButton } = prevProps;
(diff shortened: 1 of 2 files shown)


blnkt (changes requested):
Working as advertised.  

One note about the info included in the PR: the link to the Jira ticket is wrong.  Pleas update.  

And then the only other thing is that this PR doesn't include all the work as described in EPO-4668 ticket.  When we talked about this ticket this morning I noted that it included the method & variable renaming & rethinking.  But I don't see any of that work here.  Are you planing on delivering that in a separate PR?  Include it with work in a different ticket?  Skip it altogether?  Just wanna make sure I understand what to expect.

georgegomez144:
Ahhh, yes. I see the renaming bits. I'll update and push

blnkt (approved):

PR #361, opened 2021-05-04
[EPO-4699] Move all alerts into NotificationsComponent

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EPO-4699

## What this change does ##

This change moves all alerts into a central component called `NotificationsComponent`.

## Notes for reviewers ##

I had to reverse the logic in the PageContainer to return `true` if no questions are present and `true` by default. It will then return `false` if any question is unanswered. This precipitated down into the pageNav component.

## Testing ##

n/a

## Checklist ##

## Screenshots (optional) ##
n/a

### DIFF
--- a/src/components/pageNav/index.jsx
+++ b/src/components/pageNav/index.jsx
@@ -7,14 +7,14 @@ import Button from '../site/button';
 import ButtonIcon from '../site/button/ButtonIcon';
 import ArrowLeft from '../site/icons/ArrowLeft';
 import ArrowRight from '../site/icons/ArrowRight';
-import AnswerRequiredAlert from '../site/answerRequiredAlert';
-import AnswersCompletedAlert from '../site/answersCompletedAlert';
+import Notification from '../site/notifications';
 import styles from './page-nav.module.scss';
 
 class PageNav extends React.PureComponent {
   constructor(props) {
     super(props);
 
+    this.answersRequiredText = 'All Answers Are Required';
     this.state = {
       showAllRequiredAlert: false,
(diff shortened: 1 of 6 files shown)


blnkt (changes requested):
I think you misconstrued the goal of this this task.  I apologize for giving directions verbally and then not following up to ensure our expectations matched.  Re-reading your ticket description more carefully, I see now that the work you delivered is indeed "what this change does", I'm afraid the first time I read it I saw what I expected to see instead of what was there:

> This change is to create a new Notification component that will contain all the alerts within the investigation. Currently, there are two alerts floating in the Page Nav component that we wish to extract and add to a Notification component.

When I had assigned this task I had intended to communicate that the goal was to create a reusable Notification component that could be implemented in place of the two same but different notifications we already had (i.e. a Notification component that has all the shared functionality of existing notifications and has props to customize in a declarative way, and whose unique markup is the `{children}` of the component).  What you delivered is a Notifications wrapper component that houses the two notifications we already had wholesale, rather than refactoring them into something reusable.

Sorry for the miscommunication.  This work cannot be merged as is, I recommend creating a new branch off `master` and starting over altogether.  However, in the interest of time, **please do not return to this work unless all other tickets assigned to you this sprint are "in review."**  If you do not get to this ticket this sprint we will backlog it for some future free time.  If/when you do get to this ticket, feel free to reach out to discuss further the intended outcomes and motivations of this task.

georgegomez144:
I went ahead and updated the jira ticket with this information/description.

PR #364, opened 2021-05-11
[EPO-4699] Create Notification Component

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EPO-4699

## What this change does ##

This change creates the Notification component to replace the individual required and completed notifications. This new component will be able to take basic parameters and consume them to produce the desired notification.

## Notes for reviewers ##

n/a

## Testing ##

Try to continue without answering a question. This should trigger the required notifications. Then answer the questions and wait for 15s for the complete notification to show. Both should auto-hide after a few seconds. If the required notification is showing when all questions are answered it should then hide and 15s later the completed notification should show.

## Checklist ##

## Screenshots (optional) ##
n/a


blnkt (commented):
This is much more what I had in mind when I assigned this ticket.  I made a couple minor changes that're mostly just knitpicky, and two that are, I think, the right move from a maintainability and flexibility and conventional standpoint.  Please review and test and ensure it's all still working as expected.

knitpicky:
- after your refactor you ended up with quite a lot "notification" & "alert" terminology mix-matching.  I just updated names and such to go all in with notification
- variable and function names within a component are specific to that function/class so I went ahead and simplified those names to not even make reference to "notification"  so, like, `show` instead of `showNotification` since it's a moment where like "what else would it be showing?"
- use classes to control styles instead of inline styles

more flexible, maintainable, and conventional
- declare dedicated notifications when using multiple notifications: I think the difference in overhead of many of these notifications in the DOM vs swapping in an objects representing state of a single notification is negligible, and having each be its own `Notification` greatly simplifies the `showThisThing` state management in the parent component.
- using `children` instead of custom prop for notification innards: `children` is something you get for free with React and is the more conventional way to deal with markup (or other components) wrapped in a component.

georgegomez144:
Utilizing `children` makes sense to me. Wish I would have remembered that feature. Breaking the `Notification` component into two instances in the `PageNav` component is a more explicit approach. I understand why you would want that. Take care of all that extra `setState` and `timeout` stuff I was trying to use to work around some show and hide issues.

PR #365, opened 2021-05-18
[EPO-4712] Add dataset color matching in scatter plots

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EPO-4712

## What this change does ##

This change applies the same dataset color logic as the Galaxy/Hubble Plot selector widget to the galactic properties widget in order to automate the dataset color relationship.

## Notes for reviewers ##

This change primary resides in the `galactic-properties.module.scss` in more 'generic' classes (`&.inactive-#{$i}` and `&.active-#{$i}`). These two classes' styling trickles down to style their svgs, background-colors, and text according to the defined `$chartColors` found in the `_variables.scss`.

## Testing ##

n/a

## Checklist ##

### DIFF
--- a/src/components/charts/galacticProperties/galactic-properties.module.scss
+++ b/src/components/charts/galacticProperties/galactic-properties.module.scss
@@ -19,7 +19,8 @@
 
 .group-point {
   stroke: none;
-  transition: fill $timing $duration, stroke $timing $duration, opacity $timing $duration;
+  transition: fill $timing $duration, stroke $timing $duration,
+    opacity $timing $duration;
 }
 
 .invisible {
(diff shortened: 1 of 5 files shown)


src/containers/GalacticPropertiesComboContainer.jsx
```diff
@@ -99,18 +93,20 @@ class GalacticPropertiesComboContainer extends React.PureComponent {
...
+            <ScatterPlot />
             <span className="screen-reader-only">{name}</span>
           </span>
         ),
         primaryText: name,
-        className: classnames(propertyItem, {
-          [linkActive]: activeProperty === property,
+        className: classnames(style.propertyItem, {
+          [style[`inactive${i}`]]: !isActive,
+          [style[`active${i}`]]: isActive,
```
blnkt, line 109: Please reimplement where use numbered `property-item` classes (e.g. `property-item-1', `property-item-2', `property-item-3' etc.) instead of numbered `inactive` and `active`.  And then nested within each `property-item-X` class you can `&.active`.  In other words, we want like a top level class tied top these color indices instead of a bunch of these index-specific classes floating around.

src/components/charts/galacticProperties/galactic-properties.module.scss
```diff
@@ -33,72 +34,125 @@
...
+    font-weight: $medium;
+  }
 
-    .nav-avatar {
-      @include labelPrimary;
-      font-weight: $medium;
-      color: $basePrimary;
-    }
+  $i: 0;
+  @each $color in $chartColors {
```
blnkt, line 48: This is an interesting loop approach (as oppose to how we handled this in `galaxy-selector.module.scss` with a `@for` loops).  I don't know that I prefer one to the other, but please update one or the other to use the same strategy so we aren't using competing approaches to accomplish the same thing in two separate places.

blnkt (changes requested):
This is looking great but I don't think it's quite there yet:

- In your PR you write 

> This change applies the same dataset color logic as the Galaxy/Hubble Plot selector widget to the galactic properties widget in order to automate the dataset color relationship.

This is accurate to the changes in your PR, but would not seem to encompass all the work as it was described in your Jira ticket

> This change is to address how the NavDrawer icon colors should match the data point colors on the graph to help the user have a clearer connection with data. This is to say, the color of the icon in the NavDrawer (on the left of the graph) should match the plotted points in the graph.

If it is the case that the galactic properties graphs are the only nested scatterplots then this work suffices.  If there are others (was my assumption when I created this task) then the exception is to abstract this color management to a level where it can be reused across those applicable case.

- This leads me to a related point: you did a diligent job of programmatically applying colors tot he nav items, but it would appear to me that the colors in the graphs are still 100% hardcoded.  Even should this work be scoped only to the galactic properties graph (rather than all nested scatterplots) I should expect the icon color to match the point color.
- Can the work you did in your stylesheet be...consolidated with the existing approach in `galaxy-selector.module.scss` and then all carried into `nav-drawer.module.scss` instead?
- Grey is a color we should reserve for the "our data vs lsst data" graphs only (and should never represent the user data).  Please update your styles to start at the second color in `$chartColors` (i.e. skipping grey/first index).  
- Also, as noted in my Jira comment, "Color Vs Distance” graph should use one color (should not be blue or red) instead of a red -> blue gradient".  To accomplish this you can either reorder the chart colors so red and blue are not either of the first two colors, or you could run your loop from the back of the `$chartColors` list.

src/containers/GalacticPropertiesComboContainer.jsx
```diff
@@ -145,6 +141,7 @@ class GalacticPropertiesComboContainer extends React.PureComponent {
             <GalacticProperties
               className="color-brightness-vs-distance-combo"
               options={activeProperty.options}
+              color={activeProperty.color}
```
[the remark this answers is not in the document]
georgegomez144: Removed

src/containers/GalacticPropertiesComboContainer.jsx
georgegomez144: reimplemented as requested and update the scss to match

src/components/charts/galacticProperties/galactic-properties.module.scss
georgegomez144: used the @for loop method as it works nicely.

blnkt (changes requested):
I added a new commit that includes a refactor of the work you pulled into the nav-drawer stylesheet:

- I removed several nav drawer item classes which did not seem to be used in the galactic properties component (possibly artifacts of previously copy/pasting this work over from the `lightCurve` stylesheet).
- moved all nav item styles into the `&-#{i}` portion of the stylesheet.  With that syntax the style represented by the `&` is never applied and so there were some styles and a css transition that were not getting applied any longer
- consolidated the `aside` and `nav` variants under one `@for` loop since the els/classes they styled didn't seem to ever collide
- added transition to the fill on the icon so the background-color, icon fill, and label text color all shift colors on hover and click at the same rate
- renamed `.property-name` to `colorized-nav-item` in anticipation of further generalization/application to other contexts

And that last item leads me to two asks from you before closing this work out: 

1. I made a number of lil changes and tested/verified as I went, but sometimes lil changes can add up to big problems.  Please doublecheck my work and verify the graphs are still receiving the correct colors in the manner you would expect.
2. So now we have this great utility class to apply to nav-items we want to color programmatically.  Let's use it in the other context we're doing this kind of thing in: the `HubblePlotter`.  As near as I can tell with a couple minor classname changes and maybe some slight fine-tuning you can import `navStyle` into the `HubblePlotterContainer` and then delete the nav-drawer specific styles out of `hubble-plot.module.scss`.  If it looks otherwise ping me on Slack and we'll back away slowly.  But if it looks A-ok please make it so.

When the above 2 items are addressed I'll review a final time and then rebase & merge this work in.  Thank you much.

georgegomez144:
Okay, @blnkt I tested the updates and they worked as expected. I also dug through the HubblePlotter and added the classes where it was required to apply the colors and got rid of unused classes and flags. Working as expected. Triple check?

blnkt (approved):

PR #366, opened 2021-05-19
[EPO-4713] Add an SvgShape component to load dynamic shapes

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EPO-4713

## What this change does ##

This change adds the 'dynamic' SvgShape component to our scatter plots. This update effects the `galacticProperties/index.jsx`. Now the shapes default to a circle and when multiple datasets are loaded will show in this order: `[ circle, square, triangle, diamond ]`. You can also explicitly set the shape by either providing the shape name or its index as seen in the array order through the options properties:
- `svgShape: 'circle'` or whatever shape mentioned above
- `svgShapes: ['triangle', 'diamond']` shapes will be loaded in order
- 
Also in the `galacticProperties/index.jsx` you can set it through its index:
- `svgShapeIndex={0}` through 3
## Notes for reviewers ##

Pertinent changes can be found in:
`galacticProperties/index.jsx` that permeates through `Points` and `Point` into the new `shared/shapes/index.jsx`.
Also, other files of interest are the `fragments/widgetOptions.js` where two additional parameters have been added: `svgShape` and `svgShapes`. The former is a string and the latter is an array of strings that will render in the order presented.

Include an indication of how detailed a review you want on a 1-10 scale. - 6 = "Please make sure there are no obvious errors and that you believe it does what it says it does"

## Testing ##

n/a

## Checklist ##

### Before:
image
image


src/components/charts/galacticProperties/index.jsx
```diff
@@ -289,6 +289,8 @@ class GalacticProperties extends React.Component {
       options,
     } = this.props;
 
+    const { svgShape, svgShapes } = options || {};
```
blnkt, line 292: do we need both options?  Or can we just use `svgShapes` and in cases where you're only wanting one always use `svgShapes[0]`?

src/components/charts/galacticProperties/Point.jsx
```diff
@@ -1,59 +1,17 @@
...
-        .ease(d3EaseElastic)
-        .attr('r', this.baseRadius * 2);
-    } else {
-      $point
-        .transition()
-        .duration(400)
-        .attr('r', this.baseRadius);
-    }
+    this.height = 15;
+    this.width = 15;
```
blnkt, line 9: As noted elsewhere, it seems like you're always adjusting the actual height and width used in each shape.  If that must be the case, then maybe no value in passing these hardcoded values down?  Or at the very least, since all our shapes are symmetrical (...even?..."fit in a square") maybe just have one value?  And since only yused in one place can be declared in prop it's assigned to rather than as a class variable.

src/components/charts/shared/shapes/Triangle.jsx
```diff
@@ -0,0 +1,40 @@
...
+      <polygon
+        className={classes}
+        points={`${width / 2},${this.padding} ${width},${height} 0,${height}`}
+        fill={fillColor}
+        stroke={fillColor}
+        tabIndex="0"
+      />
+    );
+  }
+}
```
blnkt, line 31: Since all the other shapes used are regular polygons, I was thinking this would be an equilateral triangle.  Given the length of one side (a), the height (h) of an equilateral triangle is `h = (1/2) * √3 * a`.  I'm not sure off the top of my head what this means for the space separated list of points that define the vertices of your triangle here, but let me know if you want to puzzle through it together.

src/containers/GalacticPropertiesComboContainer.jsx
```diff
@@ -22,6 +22,7 @@ class GalacticPropertiesComboContainer extends React.PureComponent {
         name: 'Brightness vs Distance',
         color: 'rgb(254, 216, 40)',
         options: {
+          svgShape: 'triangle',
```
blnkt, line 25: Would using the `ChartSwitcherContainer` allow you to adopt your `svgShapes` options conventions more consistently here?  Feels of to go through the trouble of setting up this system of options that get passed down to the widgets and then have the widget that inspired this feature in the first place balk that new convention.

blnkt (changes requested):
Sick.  It is awesome to see all these different shapes come to live in the graphs.  I gave Ardis a sneak peak at some of the difference symbol options and she is very enthusiastic about the positive impacts your work will have on helping students achieve the desired learning outcomes.  However, there's still some odds and ends to address.  some I called out inline.  Some bigger issues I am calling out below:

As near as I can tell, all the shapes except the circle and diamond are not centered on the `x, y` coordinates:
Screen Shot 2021-05-21 at 12 12 58 PM
Screen Shot 2021-05-21 at 12 13 21 PM
Screen Shot 2021-05-21 at 12 12 37 PM
Screen Shot 2021-05-21 at 12 13 53 PM

As seen in the screenshots, to validate this suspicion I had more definitively I added to your `SvgShape` component a white 1 pixel-wide circle, and black square centered on the given `x, y` coordinates, and a black square, with the assumption that all shapes should share the same center, and fit inside the same sized square.  In case it's of use as you address this issue, that code looks like:
```
...
return (
      <>
        <rect
          height={height}
          width={width}
          x={x}
          y={y}
          fill="black"
          transform={`translate(${-width / 2} ${-height / 2})`}
        />
        <svg
          ref={this.svgEl}
          viewBox={`0 0 ${width} ${height}`}
          height={height}
          width={width}
          x={x - height / 2}
          y={y - width / 2}
        >
          <SVGShape
            {...{
              classes,
              fillColor,
              height,
              width,
            }}
          />
        </svg>
        <circle cx={x} cy={y} r="1" fill="white" stroke="white" tabIndex="0" />
      </>
    );
...
```

[As we discussed in slack](https://lsstc.slack.com/archives/DPA2M16PM/p1621458118018700), having different shapes in graphs with multiple datasets plotted in one plot is not a desired feature at this time and is outside the scope of this ticket.  After consulting Ardis, she does not think this will ever be desired behavior for these types of plots (i.e. will only ever want different shapes for sets of nested graphs).  Please update to not apply any shape but the circle to `multiple` type graphs.

Finally, I just want to make the same kind of comment I did in my review of EPO-4712: The Jira ticket proposes an outcome where all scatterplots in our investigations have this new point-shape option available, but the work delivered in this PR would only seem to impact the `galacticProperties` widget.  Just as with the color management work, I saw this ticket as an opportunity to revisit and refactor and create reusable components that could then be applied more holistically.  But if it is the case that the galacticProperties graphs are the only case where this would have applied, so be it, but please, when departing from the work as it was assigned point out in your PR description the departure and describe why the departure is sensible.  And if this isn't the only case of these nested graphs, it feels like there is probably more left work to do.

src/components/charts/galacticProperties/index.jsx
```diff
@@ -389,6 +392,8 @@ class GalacticProperties extends React.Component {
                         hoveredData,
                         offsetTop,
                       }}
+                      svgShape={multipleSvgShape}
+                      svgShapeIndex={i}
```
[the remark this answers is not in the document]
georgegomez144: Having the ability to provide a string value to choose the desired shape is great for when we're being specific about what shape we want (e.g. giving the user the option from the Ed Hub). Having the indices allows the shapes to programmatically be chosen.

src/components/charts/shared/shapes/index.jsx
```diff
@@ -0,0 +1,102 @@
...
+        .attr('x', x - height / 2 - 5)
+        .attr('y', y - width / 2 - 5)
+        .attr('height', height + 10)
+        .attr('width', width + 10);
+    } else {
+      $point
+        .transition()
+        .duration(400)
+        .attr('x', x - height / 2)
+        .attr('y', y - width / 2)
```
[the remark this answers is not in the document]
georgegomez144: I'm not 100% sure I understand how to use the transform attribute in this case.

src/components/charts/shared/shapes/Circle.jsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+class Circle extends React.PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.padding = 0.5;
```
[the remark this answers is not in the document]
georgegomez144: Each padded value is different for each shape in order to keep their sizing, more or less, the same. This is only because each shape is created and centered slightly different than the other. This is the only way I could think to address this.

src/components/charts/galacticProperties/index.jsx
georgegomez144: I tried applying this and it didn't work for me. Doesn't quite make sense to me.

blnkt: okie doke

src/components/charts/shared/shapes/Triangle.jsx
georgegomez144: Not sure I grasp the math here when plotting this in svg space.

src/containers/GalacticPropertiesComboContainer.jsx
georgegomez144: How do you believe the ChartSwiterContainer would be a good fit? I'm not sure I see your meaning or intended use?

georgegomez144:
@blnkt sorry, I'm doing my best to understand your comments and your vision behind them. Maybe we can discuss this tomorrow as well.

src/components/charts/shared/shapes/index.jsx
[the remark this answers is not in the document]
blnkt: Looks like you can't transform an svg, but you can wrap your nested svg in a group and transform that, which is convenient enough.  [So in this example](https://codepen.io/blnkt/pen/ZEebgWa) I've translated some triangles half their height and width.  And I think you always have these chapes' height and width, so I was thinking you could translate your shapes to have them always end up centered on their x, y coordinates, without needing to skew their actual x, y coordinates, thus keeping the values used to represent the data more closely aligned with the data itself.

src/components/charts/shared/shapes/Circle.jsx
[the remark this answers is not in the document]
blnkt: I sorta suspect this is what's causing the symbols to be off center.  They don't look perfect in the following screen shots but they sure look closer.  Maybe I'm not thinking about this correctly? 

Screen Shot 2021-05-24 at 5 45 58 PM
Screen Shot 2021-05-24 at 5 46 35 PM
Screen Shot 2021-05-24 at 5 46 17 PM
Screen Shot 2021-05-24 at 5 45 35 PM

src/components/charts/shared/shapes/Triangle.jsx
blnkt: Given `width` the `height = (1/2) * √3 * width` so `<polygon points="width/2,0 width,height 0,height" />`
should give you an equilateral triangle whose sides are all same length as side of the square, or radius of the circle.   I think...

src/containers/GalacticPropertiesComboContainer.jsx
blnkt: Was thinking you could then be modeling your charts in the pageData instead of hardcoded in this part of the container `Class`.  Which in turn would mean you would be impacting the charts' shapes in the same manner you've done elsewhere: options on the widget in the pagedata json.  But I don't have a sense if there's other bespoke stuff going on in `ComboContainer` that would make the jump to the more generalized `ChartSwitcherContainer` prohibitively difficult.

src/components/charts/shared/shapes/Circle.jsx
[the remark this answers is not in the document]
georgegomez144: I worked on centering each shape yesterday and come up with a good solution. I haven't pushed up that work yet.

src/components/charts/galacticProperties/Point.jsx
georgegomez144: Will use a `const dimension = 15` and apply that to the props as discussed. Will also use the `import Point from '../svgShape';` approach as well.

src/containers/GalacticPropertiesComboContainer.jsx
georgegomez144: Will pull of this hardcoded data and place into the json.

src/components/charts/shared/shapes/Triangle.jsx
georgegomez144: The equation worked as expected! Thank you for that!

src/containers/GalacticPropertiesComboContainer.jsx
georgegomez144: I look at extracting this work into the json, but was not able to figure out how to do so. Maybe I'm unaware of an options prop that takes an array of options, like an "options set" or something in the json query. I think that's why I hardcoded it in the first place.

For now, I've added `svgShapes: ["shapeName"]` into the json and updated the `this.properties` in the `GalacticPropertiesContainer` to allow for that customization.

PR #367, opened 2021-05-25
[EPO-4714] Add new 'CustomIcon' component to serve icons

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EPO-4714

## What this change does ##

This change adds a new CustomIcon component that contains all available icons including José's new Galaxy svg icon. The icons are loaded via a 'name' prop and all props passed in get passed on through to its svg. The 'icon' prop was then added into the `./fragments/widget.js` on line 7 to allow for specific customization in the json.

## Notes for reviewers ##

An example of this is found on page `characterizing-tno.json`. New component only applied to `ChartSwitcherContainer`, `GalacticPropertiesComboContainer`, and `HubblePlotterContainer` for now as a test and display.

## Testing ##

Add an `icon` prop into any json where the `ChartSwitcher` is used.

## Checklist ##


blnkt (changes requested):
Looking real good.  Couple of minor changes to take this to completion.

blnkt (approved):

PR #370, opened 2021-05-27
[EPO-4715] Add new Scientific Notation function

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EPO-4715

## What this change does ##

This change adds the `scientificNotation()` into the mix. This function will convert the value given into scientific notation through a specified sigFig. If no sigFig is specified, then it defualts to 3. If the number is greater than 1 or less than -1 and has 9 or fewer digits or if the number is between 1 and -1 and has 3 or fewer digits after the decimal the function will return that number rounded to sig figs.

## Notes for reviewers ##

Added the ability for the table cell to render the HTML if present.

## Testing ##

n/a

## Checklist ##

### Before:
image

### After:
image

### DIFF
--- a/src/components/charts/shared/StellarValue.jsx
+++ b/src/components/charts/shared/StellarValue.jsx
@@ -15,7 +15,14 @@ class StellarValue extends React.PureComponent {
 
     return (
       <>
-        {!isSvg && <span>{formatted}</span>}
+        {!isSvg &&
+          // eslint-disable-next-line no-underscore-dangle
+          (formatted.__html ? (
+            // eslint-disable-next-line react/no-danger
+            <span dangerouslySetInnerHTML={formatted} />
+          ) : (
+            <span>{formatted}</span>
+          ))}
         {isSvg && <tspan>{formatted}</tspan>}
         {formatted && <Unit type={type} isSvg={isSvg} />}
       </>
--- a/src/components/qas/questions/qaCalculator/equations/FindKineticEnergy.jsx
+++ b/src/components/qas/questions/qaCalculator/equations/FindKineticEnergy.jsx
@@ -2,7 +2,10 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import { QACalculatorIconUnit } from '../qQaCalculatorIcons.jsx';
-import { addTheCommas, toSigFigs } from '../../../../../lib/utilities.js';
+import {
+  addTheCommas,
+  scientificNotation,
+} from '../../../../../lib/utilities.js';
 import {
   color,
   equation,
@@ -54,7 +57,10 @@ export default function FindKineticEnergy(props) {
       <span className={addColorClass}>
         {kineticEnergy ? (
           <span>
-            {addTheCommas(toSigFigs(+kineticEnergy, 3))}
+            <span
+              // eslint-disable-next-line react/no-danger
+              dangerouslySetInnerHTML={scientificNotation(+kineticEnergy, 3)}
+            ></span>
             <QACalculatorIconUnit
               className={addColorClass}
               unit="kineticEnergy"
--- a/src/lib/utilities.js
+++ b/src/lib/utilities.js
@@ -238,6 +238,27 @@ export const toSigFigs = (number, precision) => {
   return String(roundedValue);
 };
 
+export const scientificNotation = (number, precision, returnHtml = true) => {
+  if (number === 0) return 0;
+
+  const [val, exp] = Number.parseFloat(number)
+    .toExponential()
+    .split('e');
+
+  const numPlaces = Math.abs(+exp);
+
+  if ((Math.abs(number) < 1 && numPlaces > 3) || numPlaces > 9) {
+    const coefficientAndBase = `${parseFloat(toSigFigs(+val, precision))} x 10`;
+
+    return returnHtml
+      ? renderDef(`${coefficientAndBase}<sup>${+exp}</sup>`)
+      : `${coefficientAndBase}^${+exp}`;
+  }
+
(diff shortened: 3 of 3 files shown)


src/lib/utilities.js
```diff
@@ -238,6 +238,35 @@ export const toSigFigs = (number, precision) => {
...
+  const afterTheDecimal = String(number).split('.');
+  const isInsideRange = number > -1 && number < 1;
+  const decimalMoreThanThreeChars =
+    afterTheDecimal[1] && afterTheDecimal[1].length > 3;
+
+  if (
+    (isOutsideRange && String(number).length > 9) ||
+    (isInsideRange && decimalMoreThanThreeChars)
+  ) {
+    const value = Number.parseFloat(number).toExponential(precision || 3);
```
blnkt, line 254: This is an excellent repurposing of js functionality that already existed on the `Float` prototype.  I think we can simplify this `scientificNotation` even more by checking what we get from `toExponential` instead of counting "places", and running all these difference checks on the input number ourselves.  Something like:
```
export const scientificNotation = (
  number,
  precision = 3,
  returnHtml = true
) => {
  if (number === 0) return 0;
  const [val, exp] = Number.parseFloat(number)
    .toExponential()
    .split('e');
  const numPlaces = Math.abs(+exp);

  // Should scientificNotationize
  if ((Math.abs(number) < 1 && numPlaces > 3) || numPlaces > 9) {
    const coefficientAndBase = `${parseFloat(toSigFigs(+val, 3))} x 10`;

    return returnHtml
      ? renderDef(`${coefficientAndBase}<sup>${+exp}</sup>`)
      : `${coefficientAndBase}^${+exp}`;
  }

  // Should sigFigIt instead
  const sigFigged = addTheCommas(toSigFigs(number, precision));
  return returnHtml ? renderDef(sigFigged) : sigFigged;
};
```

Feels simpler to me at least, whatchuthink?

If you like the above then please test it however you tested your work 'cause I'm not totally confident I'm thinking of every edge case.  I think if you wanna keep your method as is there are still a couple other lil fixes:

- the precision you're handing `toExponential` is how many digits past the decimal, not how many sig figs.  Sig figs are all important digits, not just those to the right of the decimal.  I got around this in my above example by not sigFig-ifying until constructing the string/html this method returns (i.e. letting `toExponential` run with no args).  I think this is why your method isn't always trimming trailing zeroes in the string/html output.
- In my testing, `number` was sometimes a float and `String(number).length > 9` is not working when `number` is a float because it's counting the decimal and those digits to the right of it in the length.  I got around this in the above by checking the exponent returned by `toExponential` instead of checking places/length of the input number.

src/components/qas/questions/qaCalculator/equations/FindKineticEnergy.jsx
```diff
@@ -54,7 +57,10 @@ export default function FindKineticEnergy(props) {
       <span className={addColorClass}>
         {kineticEnergy ? (
           <span>
-            {addTheCommas(toSigFigs(+kineticEnergy, 3))}
+            <span
+              // eslint-disable-next-line react/no-danger
+              dangerouslySetInnerHTML={scientificNotation(+kineticEnergy, 3)}
```
blnkt, line 62: let's get dangerous

src/components/charts/shared/StellarValue.jsx
```diff
@@ -15,7 +15,14 @@ class StellarValue extends React.PureComponent {
 
     return (
       <>
-        {!isSvg && <span>{formatted}</span>}
+        {!isSvg &&
+          // eslint-disable-next-line no-underscore-dangle
+          (formatted.__html ? (
```
blnkt, line 20: neat trick

blnkt (changes requested):
I was really inspired by your repurposing of `toExponential` and I think we can simplify this `scientificNotation` method even more by leveraging how `toExponential` works even more. 
 Instead of counting "places", and running all these difference checks on the input number ourselves can we do something like:
```
export const scientificNotation = (
  number,
  precision = 3,
  returnHtml = true
) => {
  if (number === 0) return 0;
  const [val, exp] = Number.parseFloat(number)
    .toExponential()
    .split('e');
  const numPlaces = Math.abs(+exp);

  // Should scientificNotationize
  if ((Math.abs(number) < 1 && numPlaces > 3) || numPlaces > 9) {
    const coefficientAndBase = `${parseFloat(toSigFigs(+val, 3))} x 10`;

    return returnHtml
      ? renderDef(`${coefficientAndBase}<sup>${+exp}</sup>`)
      : `${coefficientAndBase}^${+exp}`;
  }

  // Should sigFigIt instead
  const sigFigged = addTheCommas(toSigFigs(number, precision));
  return returnHtml ? renderDef(sigFigged) : sigFigged;
};
```

Feels simpler to me at least, whatchuthink?

If you like the above then please test it thoroughly.  I am not confident I considered every edge case.  I think if you wanna keep your method as is there are still a couple other lil fixes needed:

- the precision you're handing `toExponential` is how many digits past the decimal, not how many sig figs.  Sig figs are all important digits, not just those to the right of the decimal.  I got around this in my above example by not sigFig-ifying until constructing the string/html this method returns (i.e. letting `toExponential` run with no args).  I think this is why your method isn't always trimming trailing zeroes in the string/html output.
- In my testing, `number` was sometimes a float and `String(number).length > 9` is not working when `number` is a float because it's counting the decimal and those digits to the right of it in the length.  I got around this in the above by checking the exponent returned by `toExponential` instead of checking places/length of the input number.

georgegomez144:
Took some time to run through and test your suggested code since it's a much more simplified version...and to be honest, way cleaner than mine. Did my best to test as many cases as I could. Found that this did a pretty good job of catching them.

blnkt (approved):

PR #372, opened 2021-06-08
[EPO-4803] Add MathJax library

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EPO-4803

## What this change does ##

This changes adds the MathJax library into the investigations to be better render the equations.

## Notes for reviewers ##

I've only loaded the new MathJax component into the Mass calculator to begin with in order to test out the functionality.

## Testing ##

n/a

## Checklist ##

### Before:
image

### After:
image


src/lib/utilities.js
```diff
@@ -248,15 +252,21 @@ export const scientificNotation = (number, precision, returnHtml = true) => {
...
+    if (returnFormat === 'latex') {
+      return `${coefficient} \\times 10^{${+exp}}`;
+    }
+    return `${coefficient} x 10^${+exp}`;
   }
 
   const sigFigged = addTheCommas(toSigFigs(number, precision));
-  return returnHtml ? renderDef(sigFigged) : sigFigged;
+  if (returnFormat === 'html') return renderDef(sigFigged);
+  if (returnFormat === 'latex') return sigFigged;
```
blnkt, line 268: unnecessary since it is returning the saem thing it would if `returnFormat` was undefined

src/components/qas/questions/qaCalculator/equations/FindMass.jsx
```diff
@@ -1,75 +1,39 @@
...
-import { qaReviewHighlight } from './equations.module.scss';
+  addTheCommas,
+  scientificNotation,
+} from '../../../../../lib/utilities.js';
+import { renderUnit } from '../../../../charts/shared/unit/utilities.js';
+import { primary } from '../../../../../assets/stylesheets/_variables.scss';
+
+function addColorAndUnit(value, unit) {
+  return `\\textcolor{${primary}}{${value}} \\ \\mathrm{${renderUnit(unit)}}`;
+}
```
blnkt, line 15: I would expect this is gonna be something you'll use in every calculator, so let's abstract this to a utilities file that's in the same directory as your MathJax component and then you can either import the color in that file, or you can add color as an argument to `addColorAndUnit` and pass it in that way.

blnkt (changes requested):
Looking super great.  Just noted a couple of quick fixes.  Also, while I was messing around with stuff, I found that you can control the font-size of the text inside the mathjax svg splendidly using css, which should help ya out in EPO-4804.  If normal text is 16px, maybe let's try like 18px or 19px?  20px too big probably?

blnkt (approved):

PR #375, opened 2021-06-17
[EPO-4839] Add loop option for blink animation

georgegomez144:
- Ensures autoplay functionality still works
- Sets new loop prop default to `true`
- When loop is `false` animation will stop ready to start again
  from first frame

JIRA: https://jira.lsstcorp.org/browse/EPO-4839

## What this change does ##

This change adds the `loopAnimation` option in the json data and its defaultProp is set to `true`. If `loopAnimation` is set to `false`, the animation will play to the end and stop. When the play button is triggered again the animation will begin from the first frame.

## Notes for reviewers ##

n/a

## Testing ##

n/a

## Checklist ##

### DIFF
--- a/src/components/charts/galaxySelector/index.jsx
+++ b/src/components/charts/galaxySelector/index.jsx
@@ -209,8 +209,16 @@ class GalaxySelector extends React.PureComponent {
   }
 
   nextBlink = () => {
-    const { blinkCallback, images } = this.props;
-    blinkCallback(this.getBlink(images, 1));
+    const { blinkCallback, images, loop } = this.props;
+
+    const nextBlink = this.getBlink(images, 1);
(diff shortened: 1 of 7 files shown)


blnkt (approved):
Looking good.  Your defaultProp `loopAnimation: true` wasn't doing anything because the graphQL query returns null is any field is empty.  Which, for react, still counts as receiving something, and therefore the defaultProps were never kicking in.  This meant the Blinker was never looping anywhere because null is a falsey value (i.e. `!false` and `!null` are both true).  To get around this I committed a faux-pas and had it check for `loopAnimation === false`. 
 Not my favorite solution, but beats gong through all the pages that have a Blinker on them and adding `loopAnimation: true` everywhere.  I also just did a find/replace and made `loopAnimation` `loop` since that's the same terminology the VideoPlayer uses in its options.

PR #376, opened 2021-06-18
[EPO-4840] Add the Elapsed Time indicator

georgegomez144:
JIRA: https://jira.lsstcorp.org/browse/EPO-4840

## What this change does ##

This change adds the Elapsed Time indicator to the Galaxy selector widget that displays the elapsed days and hours of each alert. The indicator is its own separate component that takes two props, days and hours. 

## Notes for reviewers ##

n/a

## Testing ##

n/a

## Checklist ##

### DIFF
--- a/src/components/charts/galaxySelector/index.jsx
+++ b/src/components/charts/galaxySelector/index.jsx
@@ -10,6 +10,7 @@ import CircularProgress from 'react-md/lib/Progress/CircularProgress';
 import { arrayify } from '../../../lib/utilities.js';
 import { getAlertFromImageId } from './galaxySelectorUtilities.js';
 import Blinker from '../shared/blinker/index.jsx';
+import ElapsedTime from '../shared/elapsedTime/index.jsx';
 import Points from './Points';
 import Message from './Message';
 import Legend from '../shared/legend/index.jsx';
@@ -317,6 +318,16 @@ class GalaxySelector extends React.PureComponent {
     }));
   };
 
+  getAlertDaysAndHours = (id, alerts) => {
+    const currentAlert = getAlertFromImageId(id, alerts);
+    const dateDiff = currentAlert.date - alerts[0].date;
+
+    return {
+      days: Math.round(dateDiff),
+      hours: Math.round((24 / dateDiff) % 24),
+    };
+  };
+
   render() {
     const {
       data,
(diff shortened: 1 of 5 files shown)


src/components/charts/galaxySelector/index.jsx
```diff
@@ -414,14 +433,17 @@ class GalaxySelector extends React.PureComponent {
...
+            <>
+              <Blinker
+                images={images}
+                activeId={activeImageId}
+                playing={playing}
+                handleStartStop={this.startStopBlink}
+                handleNext={this.onNextBlink}
+                handlePrevious={this.onPreviousBlink}
+              />
+              <ElapsedTime {...this.getAlertDaysAndHours(activeImageId)} />
```
blnkt, line 445: wrap in `{alerts && <ElapsedTime ... />}` since we wouldn't wanna render this counter if there's nothing to count.  Plus, then you don't need to check to see if alerts are present in your `getAlertDaysAndHours` method

src/components/charts/galaxySelector/index.jsx
```diff
@@ -317,6 +318,24 @@ class GalaxySelector extends React.PureComponent {
...
 
+  getAlertDaysAndHours = id => {
+    const { activeGalaxy } = this.props;
+    const { alerts } = activeGalaxy || {};
+    let days = 0;
+    let hours = 0;
+
+    if (alerts) {
+      alerts.forEach(alert => {
+        if (alert.image_id === id) {
```
blnkt, line 329: I believe the `getAlertFromImageId` utility method retrieves the `alert` from the `image_id` and the `alerts`.  If so, use that instead of this loop.

src/components/charts/galaxySelector/index.jsx
```diff
@@ -317,6 +318,24 @@ class GalaxySelector extends React.PureComponent {
     }));
   };
 
+  getAlertDaysAndHours = id => {
+    const { activeGalaxy } = this.props;
+    const { alerts } = activeGalaxy || {};
```
blnkt, line 323: Aren't `alerts` already delivered as a prop?

blnkt (changes requested):

src/components/charts/galaxySelector/index.jsx
georgegomez144: Yes and are exactly the same alerts I'm grabbing out of the `activeGalaxy` prop.

PR #378, opened 2021-06-21
[EPO-4841] Add update dataset from Lauren with images

georgegomez144:
Changes include:
- Updating json data files to match structure
- Add images for Galaxy/Supernovae selectors
- Update json page files to point to new data files

JIRA: https://jira.lsstcorp.org/browse/EPO-4841

## What this change does ##

This change adds the new datasets created by Lauren to update the galaxy/supernovae selectors and light curves.

## Notes for reviewers ##

Use this [page](https://docs.google.com/presentation/d/1tJLi9SRvBGUsAPLPzsT88rF7JhuG9G6YhGFf_FouVgc/edit#slide=id.gdf6e3358d8_0_227) to double check data is rendered correctly/as expected.

## Testing ##

n/a

## Checklist ##

## Screenshots (optional) ##
n/a


blnkt (changes requested):
This page in the expanding universe breaks because it is looking for images for `ZTF20acxzkkf` that are no longer there: `pages/interpreting-hubble-plot-1_4.json`.  Or maybe are there but have different filenames now?  I'm not sure.  Please check it out and fix.

Otherwise this looks good.

georgegomez144:
The widget was looking for images that were not there anymore. Seems I deleted them when updating the images for the new datasets. I checked out to master to grab those images and reinserted them back into their folder. Widget on that page loads images as expected.

blnkt:
That's a good catch.  However, since `ZTF20acxzkkf` is one of the updated data/image sets I believe more maintainable solution is to just replace the references to the missing images with references to the new images.  This can be accomplished by replacing the `ZTF20acxzkkf` `alerts` array in the `galaxy_selector.json` with the updated one from `ZTF20acxzkkf.json`.  This way everywhere we're using this object's data we're using it in the same manner, and we only have one set of corresponding images.  It should be a quick fix.  I'll make the change and push it up so you can see what I mean.

blnkt:
Oh shoot, but then I was in "getting PR ready to merge" mode and rebased all the work down to one commit.  Sorry.  Auto-pilot.

PR #404, opened 2021-12-14
[C] EPO-5643 - Refactor QAs into QuestionBlock

alexgoff:

## What this change does ##

Previously the questions in investigations were rendered outside of the layout flow that the other block components exist in. The `QAs` component has been normalized into a `QuestionBlock` component that accepts a layout position the same as other block components. A default layout for QuestionBlock has also been created at the `PageContainer` level to closely approximate it's previous position after the `middle left` content. 

After looking through several investigations the layouts appear to match their previous position and no major layout change has occurred.

## Notes for reviewers ##

Observe that the questions appear in their expected location and function identically to how they did prior to this change. No work was done on the blocks themselves apart from creating a `QuestionBlock` component and adding the questions to the `blockGroup` object in the column layout components.

## Testing ##

Tested by comparing the investigation post-refactor to it's original state and verifying that questions appear in the expected location.

## Checklist ##

### DIFF
--- a/src/containers/PageContainer.jsx
+++ b/src/containers/PageContainer.jsx
@@ -67,6 +67,9 @@ class PageContainer extends React.PureComponent {
     return contents;
   }
 
+  getQuestions = questions =>
+    questions ? [{ questions, layout: { row: 'middle', col: 'left' } }] : null;
+
   getQuestionsWithNumbers = (questionNumbers, questions) => {
     if (!questions || !questionNumbers) return [];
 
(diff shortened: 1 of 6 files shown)


src/containers/PageContainer.jsx
```diff
@@ -67,6 +67,9 @@ class PageContainer extends React.PureComponent {
     return contents;
   }
 
+  getQuestions = questions =>
+    questions ? [{ questions, layout: { row: 'middle', col: 'left' } }] : null;
```
alexgoff, line 71: default layout for the questions if there are any

blnkt (approved):
Excellent.

PR #405, opened 2021-12-17
[C] Refactor BlocksLayout into BlocksColumn

alexgoff:
For JIRA: "EPO-5644 #IN-REVIEW #comment Refactor BlocksLayout into BlocksColumn"

## What this change does ##

Refactored `BlocksLayout` into `BlocksColumn` and move the logic from `Blocks` that decides which blocks belong in the current layout up to the column level. 

## Notes for reviewers ##

Check for correct layouts, layouts should match their prior appearance

## Testing ##

Compared layouts post-refactor to pre-refactor and confirmed that content blocks are rendering in their correct positions

## Checklist ##


src/components/page/SingleCol.jsx
```diff
@@ -1,23 +1,12 @@
...
-      questions,
-      answers,
-      tables,
-      images,
-      videos,
-      widgets,
-      checkpoints,
-      shared,
-    } = this.props;
+    const { title, questions, answers, shared, blocksGroups } = this.props;
```
alexgoff, line 9: removed unused props that are now inside of `blocksGroups`

src/components/page/blocks/BlocksColumn.jsx
```diff
@@ -0,0 +1,74 @@
...
+import Blocks from './index.jsx';
+
+function BlocksColumn({
+  getCol,
+  blocksGroups,
+  defaultLayout,
+  blockShared,
+  blockRows,
+}) {
+  const getLayout = layout => {
```
alexgoff, line 12: `getLayout` method extracted from `Blocks` to determine a content block's position using the `defaultLayout` as a fallback

src/components/page/blocks/BlocksColumn.jsx
```diff
@@ -0,0 +1,74 @@
...
+      const row = layout.row || defaultLayout.row;
+      const col = layout.col || defaultLayout.col;
+
+      return { row, col };
+    }
+
+    return defaultLayout;
+  };
+
+  const isInLayout = (layout, getRow) => {
```
alexgoff, line 23: determines if a content block's layout falls into the current `row` and `col`

src/components/page/blocks/BlocksColumn.jsx
```diff
@@ -0,0 +1,74 @@
...
+    const { row, col } = getLayout(layout);
+
+    return getCol === col && getRow === row;
+  };
+
+  const getRowBlockGroups = getRow =>
```
alexgoff, line 29: this method could use some closer examination to see if it can be simplified. What it does is first uses a reducer on `blocksGroups` to remove the block groups that have no content (i.e. are `null`). 

Then with the block groups that do have content checks to see if they are within the current layout position. Finally since `filter()` will return an empty array for non-matching content blocks, it runs a second filter to remove those block groups that have empty arrays.

src/components/page/blocks/BlocksColumn.jsx
```diff
@@ -0,0 +1,74 @@
...
+              return isInLayout(block.layout, getRow);
+            }),
+          });
+        }
+
+        return result;
+      }, [])
+      .filter(blockGroup => blockGroup.blocks.length > 0);
+
+  return blockRows.map(getRow => {
```
alexgoff, line 45: iterates through the rows (top, middle, and bottom as defaults) and extracts the block groups that match the current row and column then renders their blocks.

src/components/page/blocks/index.jsx
```diff
@@ -23,51 +23,28 @@ class Blocks extends React.PureComponent {
...
-      return null;
+      const BlockTag = this.blockTags[type];
+      const key = `${type}_${i}`;
+
+      return (
+        <BlockTag
+          key={key}
+          {...{ blockShared, block, row: getRow, col: getCol }}
+        />
+      );
```
alexgoff, line 38: `Blocks` is simplified into a presentational component that simply finds the correct block for the `type`

src/containers/PageContainer.jsx
```diff
@@ -122,24 +122,53 @@ class PageContainer extends React.PureComponent {
       questionsByPage,
     };
 
+    const blocksGroups = [
```
alexgoff, line 125: moved `blocksGroups` up to the parent container so that it only needs to be declared once.

src/components/page/TwoCol.jsx
```diff
@@ -41,95 +40,18 @@ class TwoCol extends React.PureComponent {
...
-            layout={{
-              getRow: 'bottom',
-              getCol: 'right',
-            }}
-            {...{ blocksGroups, blockShared }}
-          />
+          <BlocksColumn getCol="right" {...{ blocksGroups, blockShared }} />
           {this.isPosEmpty({ col: 'right' }, [
             ...(tables || []),
             ...(images || []),
```
blnkt, line 57: I think we actually take out this whole sense of Placeholders now.  And if we do then we don't need tables, images, widgets, etc. as props anymore either, all the way back up to the PageContainer...I think...which would be nice.  Let's breakout this additional change into its of jira ticket/PR in case it does bork something.

src/components/page/blocks/BlocksColumn.jsx
blnkt: What if we try address part of this where we're instantiating our `blocksGroups` array?  We can create a like `getBlocksGroups` method, and then have that method only return each block obj in the array if blocks of that type are on the page.  Pre-reducing so to speak.  And so if none of those obs were included, then `getRowBlockGroups` would never need to deal with null or empty array values for `blocks`, which I think would simplify this method almost to the point where it's just doing the like `isInLayout` filtering...maybe?  If we go this route, of course some careful, "ok, is there somewhere else that is depending on having those like 'empty' blockGroups".  And maybe there is, but we could probs weed those out without too much difficulty.

src/components/page/blocks/BlocksColumn.jsx
```diff
@@ -0,0 +1,74 @@
...
+    const rowBlockGroups = getRowBlockGroups(getRow);
+
+    return rowBlockGroups.map((rowBlockGroup, i) => {
+      const { type, blocks } = rowBlockGroup;
+      const key = `${type}-${getRow}-${getCol}-${i}`;
```
blnkt, line 50: While we're in this world, what do you think about renaming these `getCol` and `getRow` variable names to `col` and `row`?  I don't know if they maybe were functions at some point, but I think everywhere they're used now they're just like strings, and to me at least, every time I read `getBlah` I'm expecting it to be a function.  Could be another quick separate ticket.  Or maybe is overly fastidious and pedantic.

src/components/page/blocks/BlocksColumn.jsx
```diff
@@ -0,0 +1,73 @@
...
+    return defaultLayout;
+  };
+
+  const isInLayout = (layout, row) => {
+    const blockLayout = getLayout(layout);
+
+    return blockLayout.col === col && blockLayout.row === row;
+  };
+
+  const getRowBlockGroups = row =>
```
alexgoff, line 29: after moving the initial reducer of `blocksGroups` up to `PageContainer` this method can be squashed down to a single `.reduce()` that filters out the blocks for the current row/col and then adds it back to the `blocksGroups` for the given row/col if there are any.

src/containers/PageContainer.jsx
```diff
@@ -61,6 +61,9 @@ class PageContainer extends React.PureComponent {
     return true;
   };
 
+  getBlocksGroups = blocksGroups =>
+    blocksGroups.filter(blocksGroup => blocksGroup.blocks);
```
alexgoff, line 65: initial filtering to remove groups where the blocks are `null`

PR #411, opened 2022-01-07
[C] New page between 11 and 12 in StSS

alexgoff:
For JIRA: "EPO-5702 #IN-REVIEW #comment Add new page, update ordering"

JIRA: https://jira.lsstcorp.org/browse/EPO-5702

## What this change does ##

Changed the last TNO definition then moved page 17 to become the new page 12 and altered it's contents. Additionally re-numbered and updated links for the other "identifying groups" pages.

## Notes for reviewers ##

Review between pages 11 and 17 to verify ordering and content.

## Testing ##

How did you test this change?  Can the reviewer test this change for themselves?

If you wrote new tests in the change it's fine to just put "New tests in PR". If you haven't added any tests but you think your change is covered by existing tests you can say "Existing tests". And if you're not sure how to test it, say that.

## Checklist ##

### DIFF
--- a/src/data/pages/characterizing-comet.json
+++ b/src/data/pages/characterizing-comet.json
@@ -127,7 +127,7 @@
               "value": "Within inner Solar System (between 1.5 - 5.2 au)"
             },
             {
-              "label": "All of the orbits are located in the outer Solar System beyond the orbit of Jupiter (greater than 5.2 au).",
+              "label": "The orbit sizes are 30 au or larger, and objects typically orbit beyond Neptune.",
               "value": "In the outer Solar System"
             },
             {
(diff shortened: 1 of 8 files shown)


blnkt (approved):
Question ids need to be unique (within an investigation).  They do not need to be sequential.  I noticed you changed a number of question ids here.  I thiiiiink the only troublesome one here is `20`, which is in use on `characterizing-orbits-4` and then reused on `identifying-groups`.  There's no super straight-forward way to doublecheck which ids have already been used where.  In the past we've just sorta thrown in some 3 digit number when needing to insert a new question or whatever, do a find in the pages dir for `"id": "the 3 digit number"` and if it's the only one, then yahtzee.   I've gone ahead and added a new commit with this one minor change.

PR #422, opened 2022-01-27
[F] ReferenceBlock

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-5803

## What this change does ##

Move the reference modal into a ReferenceBlock component that can be put into the existing block layout flow. Add optional positioning to display the button either left center or right in it's block.

## Notes for reviewers ##

Check layout positioning of the modals and that there are no weird layout displays 

- 3

## Testing ##

Check the semi-major-axis page and the orbital interpretation pages in Surveying the Solar System for new modal location.

## Checklist ##


src/assets/stylesheets/components/content/_reference.scss
```diff
@@ -1,5 +1,16 @@
...
+    justify-content: flex-start;
+  }
+
+  &-right {
+    justify-content: flex-end;
+  }
+
+  &-center {
+    justify-content: center;
+  }
```
alexgoff, line 15: added positioning classes based on the `position` property in the reference modal JSON

src/components/page/blocks/BlocksColumn.jsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+// eslint-disable-next-line import/no-cycle
```
alexgoff, line 3: moving the reference content into a `ReferenceBlock` that is imported by the `Block` creates a cyclical dependency warning since modals display their content using the `BlockColumn` component.

it's not a true cyclical dependency since modals use a stripped down version of the block layout that can only display content, images, video, or tables so there is no way to load an additional ReferenceBlock and create a true circular dependency i.e. the instance of BlockColumn created by ReferenceBlock can't created a ReferenceBlock.

src/data/pages/categorizing-new-discoveries.json
```diff
@@ -25,34 +25,40 @@
...
-          [{ "content": "Jupiter" }, { "content": "5.2" }],
-          [{ "content": "Saturn" }, { "content": "9.5" }],
-          [{ "content": "Uranus" }, { "content": "19.2" }],
-          [{ "content": "Neptune" }, { "content": "30.1" }]
-        ],
-        "qaReview": false
-      }
-    ]
-  },
+  "reference": [
```
alexgoff, line 28: each of the reference JSON blocks is re-written into an array to match the existing block paradigm and allow for displaying multiple reference modals per page.

blnkt (commented):
tada!

PR #425, opened 2022-02-08
[F] Create FillableTable component

alexgoff:
For JIRA: "EPO-5828 #IN-REVIEW #comment Create fillable table component"

JIRA: https://jira.lsstcorp.org/browse/EPO-5828

## What this change does ##

Created a component that acts as a fillable table that can then be statically displayed using the ObservationsTable later.

## Notes for reviewers ##

Check the fillable table itself, saving and displaying the table contents later in the investigations, and viewing the table during the QA Review. Additionally check different inputs, deleting, and other actions.

## Testing ##

Start testing by going to /solar-system-summary/ in Surveying the Solar System and filling out the table on this page. Proceed through the next few pages and observe that the static table displays correctly. Finally, observe the table on the QA Review page.

## Checklist ##


src/components/charts/shared/StellarValueRange.jsx
```diff
@@ -1,18 +1,16 @@
...
     const { type, data } = this.props;
-    const minMax = extentFromSet(data, type);
 
     return (
       <>
-        <StellarValue type={type} value={minMax[0]} />
-        ` – `
-        <StellarValue type={type} value={minMax[1]} />
+        <StellarValue type={type} value={data[0]} />
+        <span> – </span>
```
alexgoff, line 12: I think the backticks are lingering from a template literal string, they did not look correct encapsulating the dash. 

This component was not in use and I changed some of the functionality. Previously it would extract the range from the dataset using an accessor and a helper method, now it simply uses the only two values in the range which are assumed to be the min/max. We could revert to extracting the min/max to allow passing full arrays.

src/components/charts/shared/observationsTables/ObservationsTable.jsx
```diff
@@ -7,27 +7,26 @@ import { tableTitle } from './observations-tables.module.scss';
 
 class ObservationsTable extends React.PureComponent {
   getCell(answers, cell) {
-    const { accessor, id, ids, data, content } = cell;
+    const { accessor, id, data, content, type } = cell;
```
alexgoff, line 10: `type` is added to format display for eccentricity/inclination using the existing formatting helper.

src/components/charts/shared/observationsTables/ObservationsTableCell.jsx
```diff
@@ -7,25 +7,18 @@ import StellarValue from '../StellarValue';
 import StellarValueRange from '../StellarValueRange';
 
 class ObservationsTableCell extends React.PureComponent {
-  rangifier(answerRange) {
```
alexgoff, line 10: replaced this whole "rangifier" and "answerRange" setup that was built around a bunch of conditionsl with a simple accessor type of "range".

src/components/qas/questions/qaFillableTable/FillableTableRangeInput.jsx
```diff
@@ -0,0 +1,132 @@
...
+
+class FillableTableRangeInput extends React.PureComponent {
+  constructor(props) {
+    super(props);
+
+    const { answer } = props;
+
+    this.state = {
+      minValue: answer.data[0] || undefined,
+      maxValue: answer.data[1] || undefined,
```
blnkt, line 15: It looks like this is barfing when `answer` is `null` (or maybe it's `undefined` or whatever).  You can probably get around this by using the "Optional chaining" operator `?.`, so like `answer?.data?.[0]` which should then try it, and if fails land in the other side of your `||` (so set the value as `undefined`).  

My guess is this is only happening when the user has never answered this question, but it could be the case that the `answer` prop never has value in the constructor.  If so, let me know and we can brainstorm how/when to peek the value in `componentDidMount` or `componentDidUpdate`.

blnkt (changes requested):
I’m seeing an error on http://localhost:8000/solar-system-summary (or https://deploy-preview-425--investigations.netlify.app/solar-system/solar-system-summary/ for that matter), which I think is the first page that uses the FillableTable Question type.  See my inline comments for what I suspect the issue to be.

src/components/qas/questions/qaFillableTable/FillableTableRangeInput.jsx
```diff
@@ -0,0 +1,141 @@
...
+
+    this.state = {
+      minValue: undefined,
+      maxValue: undefined,
+      hasFocus: false,
+      answerable: false,
+    };
+  }
+
+  componentDidMount = () => {
```
alexgoff, line 19: use lifecycle hooks to get the answer out of the props if it has been answered and put it into the state, otherwise leave as undefined.

blnkt (approved):

PR #431, opened 2022-02-14
[C] update orbital path colors

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-5827

## What this change does ##

Adds utility constants to hold the orbital body and path colors

## Notes for reviewers ##

Some small updates to the Orbital Viewer sub-components 

- 2

## Testing ##

Observe the orbital viewer and check that the orbital bodies and paths have been updated and display well.

## Checklist ##

### DIFF
--- a/src/components/visualizations/orbitalViewer/orbitalUtilities.js
+++ b/src/components/visualizations/orbitalViewer/orbitalUtilities.js
@@ -3,6 +3,42 @@ import * as THREE from 'three';
 export const AU_TO_VIZ_SCALER = 100;
 export const DAY_PER_VIZ_SEC = 1;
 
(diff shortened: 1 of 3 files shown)


src/components/visualizations/orbitalViewer/orbitalUtilities.js
```diff
@@ -3,15 +3,51 @@ import * as THREE from 'three';
 export const AU_TO_VIZ_SCALER = 100;
 export const DAY_PER_VIZ_SEC = 1;
 
+export const ORBITAL_COLORS = {
```
alexgoff, line 6: created a config object that stores the object and orbital colors for each orbital object

PR #433, opened 2022-02-17
[C] Add edit icon to fillable table

alexgoff:
For JIRA: "EPO-5884 #IN-REVIEW #comment add edit icon to fillable table"

JIRA: https://jira.lsstcorp.org/browse/EPO-5884

## What this change does ##

Add a button to activate editing the fillable table inputs when clicked as demonstrated on W3: https://www.w3.org/TR/wai-aria-practices-1.1/examples/grid/dataGrids.html#ex2_label

## Notes for reviewers ##

## Testing ##

Start at `/solar-system-summary` in Surveying the Solar System and observe the static display in the table cells. Click the icon or the text itself to activate editing. Edit contents and refresh or move forwards then backwards to the page and observe that it has reset to static display. Tab through and observe focus shift to each editable cell.

## Checklist ##

### Before: 
image
### After:
image


src/components/qas/questions/qaFillableTable/FillableTableRangeInput.jsx
```diff
@@ -90,48 +99,58 @@ class FillableTableRangeInput extends React.PureComponent {
...
+                defaultValue={answered ? answer.data[1] : ''}
+                lineDirection="center"
+                onBlur={this.handleBlur}
+                onFocus={this.handleFocus}
+                onChange={this.handleMaxChange}
+                fullWidth={false}
+              />
+            </>
+          )}
+          {!qaReview && !editing && (
```
alexgoff, line 135: added `editing` to the component state, if not currently editing then this button styled to appear as plaintext with an icon is displayed. On click the `handleEdit` method switches the state to editing mode and the text input is shown.

blnkt (changes requested):
Getting an error on the frontend when hitting any page for the first time with no Answers (see screenshot of error output in Inspector).
Screen Shot 2022-02-21 at 1 41 22 PM

Because when the page first loads `answer` can be `null` (or `undefined` or whatever) any operations using `answer` need to be behind a conditional.  Most of the uses of `answer` in `FillableTextInput` and `FillableTableRangeInput` are on the other side of conditions using the value of `answered`, but not all of them.  I think with just that change should be good to go, but could also need to use the conditional chaining in some cases?

blnkt:
Also please address the stylelint warnings.  There's also a `warn "export 'RGBFormat' was not found in 'three'` that seems to point to the `FillableTable` but I'm super not sure how can be since three.js isn't used in your component...

blnkt (changes requested):
This is working as advertised, but I would like to see us take it one more step further towards the kind of behavior we see in [that w3 example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/grid/dataGrids.html#ex2_label).  In particular, I think toggling the visibility of the "button table cell" and its edit icon `onFocus` and `onBlur` would feel a lot better.  I know, as a rule, one isn't suppose to change focus for the user, but in this case I think you would have to (changing focus from the button to the input `onClick`).  This should be accomplishable by using refs.

So I would like you to update this PR so that: 
1.  When the user clicks on the "table cell" it brings focus to the input (or the first input in the case of ranges)
2. When a "table cell" has focus the edit icon is not visible
3. When a "table cell" blurs the edit icon is visible again

That's my 2 cents at least.  Let me know if you feel otherwise, or if Ardis or Jose were adamant about it working as we see it working currently.

blnkt (approved):
image

Feels right.  Thanks for going the extra mile on this one.  I feel like this last focus the focus addition really made a big difference in terms of UX.

src/components/qas/questions/qaFillableTable/FillableTableRangeInput.jsx
```diff
@@ -41,14 +48,47 @@ class FillableTableRangeInput extends React.PureComponent {
...
       () => {
+        const { hasFocus } = this.state;
         if (focusCallback) {
-          const { hasFocus } = this.state;
           focusCallback(hasFocus);
         }
+
+        clearTimeout(this.focusTimeout);
+
+        this.focusTimeout = setTimeout(() => {
```
alexgoff, line 58: this use of setTimeout is a little hacky but it's in place because the button should only switch back if both range inputs have been blurred and if a user clicks the other input or tabs between them and then focus is checked then there is a gap where focus is lost before the next focus event fires. By using setTimeout in the setState callback we wait until after the first state operation and then wait a tick which allows focus to move to the next field and return `hasFocus` to `true`

PR #440, opened 2022-02-24
[C] Add layout options to questions

alexgoff:
For JIRA: "EPO-5885 #IN-REVIEW #comment add layout options to questions"

JIRA: https://jira.lsstcorp.org/browse/EPO-5885

## What this change does ##

Add the ability to provide a layout for individual questions 

## Notes for reviewers ##

For all pages except for /history-solar-system there should be no change to the question order. For /history-solar-system the question on the page should be moved to the bottom right. 

- 3

## Testing ##

Test on frontend and observe that there is no visual change to questions apart from /history-solar-system where a question has been moved. Experiment with adding layout to individual questions. 

## Checklist ##

### DIFF
--- a/src/components/page/blocks/QuestionBlock.jsx
+++ b/src/components/page/blocks/QuestionBlock.jsx
@@ -20,7 +20,10 @@ class QuestionBlock extends React.PureComponent {
 
   render() {
     const { blockShared, block } = this.props;
-    const { questions } = block;
+    const { question: q, number, tables } = block;
+    const primeQ = q[0];
+    const { id, questionType, qaReview, showUserAnswer } = primeQ;
+    const key = `qa-${id}`;
 
     const {
       activeQuestionId,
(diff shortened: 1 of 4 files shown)


src/components/page/blocks/QuestionBlock.jsx
```diff
@@ -31,50 +34,43 @@ class QuestionBlock extends React.PureComponent {
...
-    return (
-      <div className="qas">
-        {questions.map(question => {
-          const { question: q, number, tables } = question;
-          const primeQ = q[0];
-          const { id, questionType, qaReview, showUserAnswer } = primeQ;
-          const key = `qa-${id}`;
-
-          if (qaReviewPage && !qaReview) return false;
+    if (qaReviewPage && !qaReview) return false;
```
alexgoff, line 37: the major change is in `QuestionBlock`, previously all questions were passed to `QuestionBlock` and then mapped into individual questions, now individual questions are passed to the block and displayed.

src/lib/utilities.js
```diff
@@ -63,7 +64,15 @@ export const getContents = (content, contents) => {
 };
 
 export const getQuestions = questions =>
-  questions ? [{ questions, layout: { row: 'middle', col: 'left' } }] : null;
+  questions
```
alexgoff, line 67: instead of providing a `questions` object that contains an array of questions and a single layout, each question now has a layout embedded into the question object and the `questions` is an array of `question` objects which is in line with how other page content is managed.

this is because in the JSON questions are represented already by an array and so to add a top-level layout to all of the questions would require re-writing ~150 JSON files. An added benefit is that we can add a layout to individual questions instead of positioning just the entire block of questions.

blnkt (approved):

PR #447, opened 2022-03-04
Section breaks

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-5886

## What this change does ##

Add functionality for section breaks and section breaks for Surveying the Solar System. 

## Notes for reviewers ##

This change is fairly extensive in Surveying the Solar System but should be lower impact to other investigations. The only change for other investigations should be that all pages will be lumped into a single "section" that will remove the checkpoint icons from the progress bar.

In Surveying the Solar System there should be 3 sections displayed along with 2 new progress update pages indicating the end of sections 1 and 2.

- 6

## Testing ##

Test Surveying the Solar System and another non-sectioned investigation (Coloring the Universe). Check

- Progress bars render correctly including correct widths for number of pages per section
- Progress bar percentage displays correctly
- Progress bar markers transition correctly between incomplete and complete state
- New section pages are inserted at the correct location and can be traversed forwards and backwards.

## Checklist ##


gatsby-node.js
```diff
@@ -156,12 +156,13 @@ exports.createPages = async ({ graphql, actions }) => {
       query {
         allPagesJson(
           filter: { investigation: { eq: "solar-system" } }
-          sort: { fields: [order, investigation], order: ASC }
+          sort: { fields: [sectionId, order, investigation], order: ASC }
```
alexgoff, line 159: Added sectionId as the primary sort field, will fall back to order. This maintains backward compatibility (un-sectioned investigations will be sorted by order) and simplifies order so pages are not placed in relation to the entire pageset, but just to their own section.

src/assets/stylesheets/components/site/progress/_header-progress.scss
```diff
@@ -7,6 +7,72 @@
   height: $siteProgressHeight;
   padding: 0 $minPadding;
   background-color: rgb(245, 245, 245);
+
```
alexgoff, line 10: for the header progress bar, create an alternating set of colors.

src/assets/stylesheets/components/site/progress/_progress-linear.scss
```diff
@@ -31,6 +31,10 @@
   border-radius: $progressLinearMarkerWidth / 2;
   transform: translateX(-($progressLinearMarkerWidth / 2));
 
+  &.filled {
```
alexgoff, line 34: new "filled" marker

src/components/page/sectionBreak.jsx
```diff
@@ -0,0 +1,37 @@
...
+import PropTypes from 'prop-types';
+import ImageLoader from '../site/imageLoader';
+import Block from './blocks/index';
+
+class SectionBreak extends React.PureComponent {
+  render() {
+    const { title, blocksGroups, shared: blockShared } = this.props;
+    const sectionBreakImagePath = '/images/section_break_celebration.gif';
+
+    return (
```
alexgoff, line 11: basic page for section breaks. This builds on content blocks and is similar to the singleCol layout but simplified and can be customized easily.

src/components/site/header/HeaderProgress.jsx
```diff
@@ -1,40 +1,63 @@
 import PropTypes from 'prop-types';
 import React from 'react';
+import last from 'lodash/last';
 import CustomIcon from '../icons/CustomIcon';
 import LinearProgress from '../linearProgress';
 import LinearProgressMarker from '../linearProgress/LinearProgressMarker';
 
 class HeaderProgress extends React.PureComponent {
+  getDisplayValue = (currentPage, totalPages) =>
```
alexgoff, line 9: `displayValue` will show the progress percentage in relation to the entire investigation, even when the progress marker is inside of a specific section's progress bar.

src/components/site/header/HeaderProgress.jsx
```diff
@@ -1,40 +1,63 @@
...
+    currentPage ? Math.round((currentPage / totalPages) * 100) : 0;
+
+  getProgressValue = (currentPage, section) => {
```
alexgoff, line 12: `progressValue` is the progress within a specific section

src/components/site/header/HeaderProgress.jsx
```diff
@@ -1,40 +1,63 @@
...
+                value={this.getProgressValue(pageNumber, section)}
+                displayValue={this.getDisplayValue(pageNumber, totalPages)}
+                style={{ width: `${(section.length / totalPages) * 100}%` }}
+              >
+                {lastPage !== totalPages && (
+                  <LinearProgressMarker
+                    filled
+                    completed={pageNumber >= lastPage}
+                    progress={100}
+                    style={{ zIndex: pageNumber === lastPage ? -1 : 1 }}
```
alexgoff, line 45: set the z-index if the progress marker overlaps with the ending marker so that the progress marker displays on top. The progress marker cannot default display on top because after a progress bar is completed the marker is left at 100% but hidden by the ending marker.

src/layouts/index.jsx
```diff
@@ -51,6 +51,25 @@ class Layout extends React.Component {
     }, []);
   };
 
+  getSections = () => {
+    const { pages } = this.state;
+    const sections = [];
```
alexgoff, line 56: populate the pages in each section, if a sectionId is not specified they will default to section 0 which maintains backward compatibility with un-sectioned investigations

src/layouts/index.jsx
```diff
@@ -132,6 +151,7 @@ class Layout extends React.Component {
       totalQAsByPage: this.getTotalQAsByPage(),
       questionNumbersByPage: this.getQuestionNumbersByPage(),
       checkpoints: this.getCheckpoints(),
```
alexgoff, line 153: checkpoints have sort of fallen by the wayside with the introduction of sections. I am open to removing the checkpoints code completely since it is now unused.

src/components/page/sectionBreak.jsx
```diff
@@ -0,0 +1,37 @@
...
+      <div className="section-break">
+        <ImageLoader
+          src={sectionBreakImagePath}
+          alt="Animated image of fireworks"
+        />
```
blnkt, line 16: The fireworks gif img needs to be set to `width: 100%`.  As is, it is overflowing as soon as its container is narrower than its max width.

blnkt (changes requested):
Working as advertised.  I think your implementation all makes sense and its got all the shiny features.  Just noticed 1 small styling bug.

src/components/page/sectionBreak.jsx
alexgoff: added `image-responsive` class that sets:

```
max-width: 100%;
height: auto;
```

and a prop to set it in `ImageLoader`

PR #453, opened 2022-03-21
[C] Change QaReview sorting to sectionId primary

alexgoff:
For git: "Resolves EPO-6062"
For JIRA: "EPO-6062 #IN-REVIEW #comment Make sectionId primary sort"

JIRA: https://jira.lsstcorp.org/browse/EPO-6062

## What this change does ##

Remove the secondary sorting in QAReviewContainer and use graphQL to sort by sectionId then order, add sectionId sorting to the gatby node build

## Notes for reviewers ##

Check that review questions display in correct order.

- 2

## Testing ##

Can test by going to the QAReview page and verifying question order

## Checklist ##

### DIFF
--- a/src/containers/QAReviewContainer.jsx
+++ b/src/containers/QAReviewContainer.jsx
@@ -122,11 +122,8 @@ class QAReviewContainer extends React.PureComponent {
     return rw;
   }
 
-  getOrderedPages(pages, id, investigation) {
-    return filter(pages, ['investigation', id || investigation]).sort(
-      (a, b) => a.order - b.order
-    );
-  }
+  getOrderedPages = (pages, id, investigation) =>
+    filter(pages, ['investigation', id || investigation]);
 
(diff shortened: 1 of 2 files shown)


src/containers/QAReviewContainer.jsx
```diff
@@ -122,11 +122,8 @@ class QAReviewContainer extends React.PureComponent {
     return rw;
   }
 
-  getOrderedPages(pages, id, investigation) {
-    return filter(pages, ['investigation', id || investigation]).sort(
-      (a, b) => a.order - b.order
```
alexgoff, line 127: this method was performing a redundant sort on the pages, they are already sorted by the `order` property from graphQL and so after `sectionId` was added as the primary sort, this sort would undo that and revert it to sorting by `order`

blnkt (approved):

PR #457, opened 2022-03-24
[C] set up URL routing for i18n

alexgoff:
For git: "Resolves EPO-6066"
For JIRA: "EPO-6066 #IN-REVIEW #comment set up URL routing for i18n"

JIRA: https://jira.lsstcorp.org/browse/EPO-6066

## What this change does ##

Set up i18next to add URL path routing for locales. Redirect internal links to the appropriate URL locale. Set up fallback language for missing translations.

## Notes for reviewers ##

This change will allow switching between English and Spanish using the URL path. Only one Spanish translation is currently in place (table of contents -> Tabla de contenido) and the remaining translation keys should display their English fallback.

## Testing ##

There are a couple ways to test this, some thought about the order of detecting the user's language may crop up here as well. i18n uses this default order to detect locale:

'querystring', 'cookie', 'localStorage', 'sessionStorage', 'navigator', 'htmlTag'
 
Prefixing a path with `es/` should be enough to override the locale stored locally and change the language from English to Spanish. Alternatively the local storage can be cleared and the browser language changed. This raises the question though, should browser language (navigator) take priority in the detection order over local/session/cookie storage?

Should be able to:

- click a link and be taken to a page with the same locale you are currently on
- enter a new locale into the URL path and load a different locale
- click a link and be taken to the new locale after changing the URL path
- remove the locale from the URL path and have the default locale (EN) displayed
- use a locale and show English translations for any missing keys

## Checklist ##

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -29,12 +29,13 @@ module.exports = {
       resolve: `gatsby-plugin-react-i18next`,
       options: {
         localeJsonSourceName: `locale`,
-        languages: [`en`, `es`],
-        defaultLanguage: `en`,
-        langKeyDefault: 'en',
+        languages: ['en', 'es'],
+        defaultLanguage: 'en',
         useLangKeyLayout: false,
+        prefixDefault: false,
         i18nextOptions: {
           debug: false,
+          fallbackLng: 'en',
           interpolation: {
             skipOnVariables: false,
             escapeValue: false, // not needed for react as it escapes by default
--- a/src/components/pageNav/index.jsx
+++ b/src/components/pageNav/index.jsx
@@ -1,8 +1,7 @@
 /* eslint-disable react/jsx-handler-names */
 import React from 'reactn';
 import PropTypes from 'prop-types';
-import { Link } from 'gatsby';
-import { Trans, withTranslation } from 'gatsby-plugin-react-i18next';
+import { Trans, withTranslation, Link } from 'gatsby-plugin-react-i18next';
 import classnames from 'classnames';
 import Button from '../site/button';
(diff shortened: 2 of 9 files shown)


gatsby-config.js
```diff
@@ -29,12 +29,13 @@ module.exports = {
       resolve: `gatsby-plugin-react-i18next`,
       options: {
         localeJsonSourceName: `locale`,
```
alexgoff, line 31: a couple changes necessary to the gatsby plugin config and i18next's options

src/components/pageNav/index.jsx
```diff
@@ -1,8 +1,7 @@
 /* eslint-disable react/jsx-handler-names */
 import React from 'reactn';
 import PropTypes from 'prop-types';
-import { Link } from 'gatsby';
-import { Trans, withTranslation } from 'gatsby-plugin-react-i18next';
+import { Trans, withTranslation, Link } from 'gatsby-plugin-react-i18next';
```
alexgoff, line 4: all `Link` components are now loaded from the i18next plugin so that they redirect to the appropriate locale path

src/containers/EndingContainer.jsx
```diff
@@ -185,8 +186,8 @@ class LastPage extends React.PureComponent {
 export default LastPage;
 
 export const query = graphql`
-  query($language: String!) {
-    locales: allLocale(filter: { language: { eq: $language } }) {
+  query {
```
alexgoff, line 189: the `$language` filter is removed from all graphql queries. If it is used then only the current locale will be loaded and fallback keys will fail. There is perhaps a way we could alter this to load both the current locale plus the default locale so it does not query every single language available.

PR #458, opened 2022-04-01
[B] Fix section breaks

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-6125

## What this change does ##

Fix section breaks with new explicit data structure replacing the previous derived data structure

## Notes for reviewers ##

Change can be reviewed on the frontend using this branch or on branch `epo-6123-6125-combined` to review with the fixes from EPO-6123. Check that section breaks work on investigations where sections have been configured (solar system) and those without sections (coloring-universe)

## Testing ##

Clear local storage, begin by testing an investigation with sections (solar system) and check that section breaks correctly render in the table of contents. Check that pages can be traversed without console errors, including the QA review page. Verify all of this on dev and production environments.

Prod environment can be built locally with:

`netlify build --context solar-system-staging`
`netlify dev -d public`

other contexts are available in `netlify.toml` 

## Checklist ##


gatsby-node.js
```diff
@@ -60,10 +63,10 @@ exports.createPages = async ({ graphql, actions }) => {
       query {
         allPagesJson(
           filter: { investigation: { eq: "exploding-stars" } }
-          sort: { fields: [sectionId, order, investigation], order: ASC }
+          sort: { fields: [sectionOrder, order, investigation], order: ASC }
```
alexgoff, line 66: "sectionId" has been replaced with a simple "sectionOrder" which then maps to an array of available sections.

src/components/site/header/index.jsx
```diff
@@ -8,7 +8,7 @@ import Close from '../icons/Close';
 import Menu from '../icons/Menu';
 
 import styles from './header.module.scss';
-import HeaderProgress from './HeaderProgress';
+import HeaderProgressContainer from '../../../containers/HeaderProgressContainer';
```
alexgoff, line 11: renamed this to a container since that is what it's function is in line with

src/data/pages/big-view.json
```diff
@@ -1,7 +1,7 @@
 {
   "id": "solarsystem07",
   "investigation": "solar-system",
-  "sectionId": "1",
+  "sectionOrder": 0,
```
alexgoff, line 4: renamed these properties and changed them to start from 0 instead of 1

src/layouts/index.jsx
```diff
@@ -51,19 +49,26 @@ class Layout extends React.Component {
...
 
-  getSections = () => {
+  getSections = id => {
     const { pages } = this.state;
-    const sections = [];
+    const { investigations } = this.props;
+    const investigation = find(investigations, { id });
+    const defaultSection = [{ sectionName: 'default', pages: [] }];
+    const sections =
+      get(investigation, 'sections', defaultSection) || defaultSection;
```
alexgoff, line 58: updated the logic in here, `investigation` is undefined in one prod build case for the 404 page which created exceptions trying to get `sections` out of it. Lodash `get` provides the option for a default fallback and protects against undefined exceptions, the `||` operator then handles cases where `sections` is null.

blnkt (approved):

PR #461, opened 2022-04-04
[C] Add orbital randomization

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-6065

## What this change does ##

Adds the capability to pick a random source from a dataset in the orbit viewer and preserve it using the global store

## Notes for reviewers ##

Change is limited to the `OrbitViewerContainer` component and then the 3 "classifying objects" pages

- 3

## Testing ##

Test by visiting classifying-objects, classifying-objects-1, and classifying-objects-2 and checking that each displays an individual orbit. Revisit each page and confirm that the same orbit is displayed.

Clear the local storage and reload the application, verify that each page now displays a different orbit (it's possible one could display the same orbit again but the odds are very very low, if each one shows the same orbit you are either in danger of winning the lottery or it's not working)

## Checklist ##

### DIFF
--- a/src/containers/OrbitalViewerContainer.jsx
+++ b/src/containers/OrbitalViewerContainer.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
+import { randomIntFromInterval } from '../lib/utilities.js';
 import API from '../lib/API.js';
 import ConditionalWrapper from '../components/ConditionalWrapper';
 import NavDrawer from '../components/charts/shared/navDrawer/index.jsx';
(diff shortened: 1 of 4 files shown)


src/containers/OrbitalViewerContainer.jsx
```diff
@@ -74,6 +77,27 @@ class OrbitalViewerContainer extends React.PureComponent {
     }
   }
 
+  getSetOrbitData = (response, randomSource) => {
+    const { data } = response;
+    const { answers, updateAnswer } = this.props;
+    const { pageId } = this.global;
+
+    if (randomSource) {
```
alexgoff, line 85: if `randomSource` is set, generate a unique key for the orbit viewer on this page. 

One limitation here is that the keys are by pageId so if there are multiple orbit viewers on a page only one random orbit can be persisted. 

The only alternative I can think of is to start putting an ID in the widget and using that as the lookup. Thoughts on if this is worth the time to add? Maybe add an ID and use the page key as a backup?

PR #466, opened 2022-04-13
[C] add StSS i18n code support

alexgoff:
For git: "Resolves EPO-6126"
For JIRA: "EPO-6126 #IN-REVIEW #comment add code support for i18n"

JIRA: https://jira.lsstcorp.org/browse/EPO-6126

## What this change does ##

Add code support to translate all of the content in Surveying the Solar System. Went page-by-page in Surveying the Solar System and checked on all content blocks, questions, and widgets that were used then added the necessary translation components or methods for each. With this change there should be 100% coverage for key-value translations in Surveying the Solar System

## Notes for reviewers ##

The change is mostly in components used by investigations and in areas of Surveying the Solar System that were converted to key-value pairs to test the changes. It should be tested both in Surveying the Solar System EN/ES versions and in non-translated investigations to check that work does not break any untranslated areas.

## Testing ##

Test using the deployment in this PR and first check that Surveying the Solar System can switch between English and Spanish and that the pages are rendering correctly. English should be unchanged and the Spanish pages will be mostly-English. Check other investigations to verify that their English pages are still functioning.

## Checklist ##


gatsby-config.js
```diff
@@ -42,6 +42,23 @@ module.exports = {
           },
           keySeparator: '.',
           nsSeparator: '::',
+          react: {
```
alexgoff, line 45: added basic support for parsing HTML elements beyond what is already included by default. Additional support is added in areas like `ContentBlock` where there are nested elements which i18next does not support out of the box.

src/components/charts/shared/Tooltip.jsx
```diff
@@ -85,9 +87,13 @@ class Tooltip extends React.PureComponent {
   renderValue(accessor, data, unit) {
     return (
       <>
-        <span>{getValue(accessor, data)}</span>
+        <span>
+          <TranslatedValue {...{ accessor, data }} />
```
alexgoff, line 91: switched to using the `TranslatedValue` component instead of the `getValue` utility method.

src/components/page/blocks/ContentBlock.jsx
```diff
@@ -24,12 +25,26 @@ class ContentBlock extends React.PureComponent {
   render() {
     const { block, row } = this.props;
     const { content } = block || {};
+    const transComponents = {
```
alexgoff, line 28: mapping for basic HTML elements that need to be parsed. As we load in more content from other investigations this may grow. This is mostly so that we can support nested content which i18next will not parse using just the config-supported HTML elements.

src/components/site/forms/table/index.jsx
```diff
@@ -56,7 +57,7 @@ class Table extends React.PureComponent {
                           'row-title': includeRowTitles && j === 0,
                         })}
                       >
-                        {col}
+                        <Trans>{col}</Trans>
```
alexgoff, line 60: this can result in an attempted double translation, some child content in `col` is just plaintext that has been unprocessed but in other instances it is wrapped in `<span>` tags at the `ObservationTable` level. There's no real risk to trying to parse it twice but if debugging is turned on it will throw a key-not-found warning.

src/components/site/utilities/TranslatedValue/index.jsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { useTranslation } from 'gatsby-plugin-react-i18next';
+import { isArray } from 'lodash';
+import { getValue, addTheCommas, formatValue } from '../../../../lib/utilities';
+
+const TranslatedValue = ({ accessor, data }) => {
```
alexgoff, line 7: mirror of the `getValue` utility but utilizing React and adding support for translations. Will fall back on `getValue` using the same logic that `getValue` uses to fall back on returning the raw data.

src/layouts/index.jsx
```diff
@@ -181,26 +182,38 @@ class Layout extends React.Component {
...
-        />
-        {investigation && (
-          <TableOfContents
-            visible={tocIsOpen}
-            toggleToc={this.toggleToc}
-            investigation={investigation}
-            isAll={!envInvestigation || envInvestigation === 'all'}
-            pages={pages}
-          />
+      <Translation>
```
alexgoff, line 185: to get the investigation title translated this root layout was wrapped in the `Translation` component to give it access to the `t` method. The normal HOC pattern could not be applied here because this page is already wrapped in the `StaticQuery` HOC which doesn't use normal children but uses a `render` method that won't support the `withTranslation` HOC

PR #467, opened 2022-04-22
[C] move StSS content to key-value pairs for i18n

alexgoff:
For git: "Resolves EPO-6204"
For JIRA: "EPO-6204 #IN-REVIEW #comment move StSS content to key-values"

JIRA: https://jira.lsstcorp.org/browse/EPO-6204

## What this change does ##

Moves all of the content for Surveying the Solar System out of it's page JSON files and into translation files. Additionally makes a handful of code changes to better translation and avoid double-translating content.

## Notes for reviewers ##

This change impacts every page of Surveying the Solar System although it should be an invisible change, the English content should not be visibly different in any way. 

- 6

## Testing ##

Test using the deployment attached to this pull request, clear the local cache first. A simple test would be to visit each page of the investigation and look for obviously missing or untranslated content. 

An additional test can be done by running the code locally and in `gatsby-config.js` setting `debug: true` in the i18next configuration and going through each page and checking for debug messages indicating a key does not have a value. There should be a small number of keys that will display, mostly these are rows of tables that are attempted to be translated but have not been put in key-value pairs because they are not localizable (numbers)

## Checklist ##

### DIFF
--- a/src/components/charts/shared/observationsTables/ObservationsTable.jsx
+++ b/src/components/charts/shared/observationsTables/ObservationsTable.jsx
@@ -67,7 +67,7 @@ class ObservationsTable extends React.PureComponent {
   translateCells = cells => {
     const { t } = this.props;
 
-    return cells.map(t);
+    return cells ? cells.map(t) : cells;
   };
 
   render() {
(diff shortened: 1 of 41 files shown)


src/components/charts/shared/observationsTables/ObservationsTable.jsx
```diff
@@ -71,9 +82,9 @@ class ObservationsTable extends React.PureComponent {
         )}
         <Table
           className="observations-table"
-          colTitles={colTitles}
+          colTitles={this.translateCells(colTitles)}
```
alexgoff, line 85: translation has been moved entirely out of low level components like `Table` and is instead performed in it's wrappers like `ObservationsTable` to avoid duplicate translation and mixing translation code with unrelated code.

PR #473, opened 2022-05-02
[C] reset global state when clearing answers

alexgoff:
For git: "Resolves EPO-6247"
For JIRA: "EPO-6247 #IN-REVIEW #comment reset global state when clearing"

JIRA: https://jira.lsstcorp.org/browse/EPO-6247

## What this change does ##

Resolves a bug where when resetting the global state, the callback updates the local storage with the global state pre-reset. This would result in not being able to clean out old code stored locally. Instead reactn's built-in reset tool is used and the callbacks and reducers are reset at the same time.

## Notes for reviewers ##

Where should the reviewer look for the change?  Do they review the change on the frontend?  How extensive is the change?  Does it affect Components used in a variety of contexts?  Anything you'd like reviewers to pay particular attention to or questions you have about the change.

- 2

## Testing ##

This change can be tested by adding a new property to the local storage ex. `bogus: true`, clicking "clear all answers" and then checking that the local storage no longer has the added property i.e. it matches the empty state defined in the application

## Checklist ##

### DIFF
--- a/src/layouts/index.jsx
+++ b/src/layouts/index.jsx
@@ -38,18 +38,6 @@ class Layout extends React.Component {
     this.store.addReducers();
   }
 
-  getCheckpoints = () => {
-    const { pages } = this.state;
-
-    return pages.reduce((checkpoints, page) => {
-      if (page.checkpoints && page.checkpoints.length) {
-        checkpoints.push(page.pageNumber);
-      }
-
-      return checkpoints;
-    }, []);
-  };
-
   getSections = id => {
     const { pages } = this.state;
     const { investigations } = this.props;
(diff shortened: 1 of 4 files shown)


src/layouts/index.jsx
```diff
@@ -38,18 +38,6 @@ class Layout extends React.Component {
     this.store.addReducers();
   }
 
-  getCheckpoints = () => {
```
alexgoff, line 41: this was written for the progress bar but then went unused, removing it

src/state/GlobalStore.js
```diff
@@ -43,6 +41,10 @@ class GlobalStore {
 
   addReducers() {
     addReducer('empty', () => {
+      resetGlobal();
```
alexgoff, line 44: the global callback that updates local storage was not receiving the reset global state, perhaps the async operations were occurring in a different order than expected. `resetGlobal()` completely clears the state, then callbacks and reducers are re-added and the fresh global state is returned to repopulate the state. This triggers the correct order of saving the state to local storage after it has been cleared.

PR #490, opened 2022-10-27
[C] update to styling and sectionbreaks

alexgoff:
For JIRA: "EPO-7155 #IN-REVIEW #comment update to styling and sectionbreaks"

JIRA: https://jira.lsstcorp.org/browse/EPO-7155

## What this change does ##

- Use finalized icon
- Update second progress bar color
- Add a variation on the section break GIF for the final section break in an investigation

## Notes for reviewers ##

Look for changes on the frontend deployment

## Testing ##

- Verifying favicon is difficult since favicons are heavily cached
- Verify that the progress bar displays a turquoise as its second color instead of red
- Verify that the final section break has a different image that is available in English and Spanish and other section breaks are unchanged

## Checklist ##


src/components/page/sectionBreak.jsx
```diff
@@ -7,7 +7,14 @@ import Block from './blocks/index';
 class SectionBreak extends React.PureComponent {
   render() {
     const { blocksGroups, shared: blockShared, t } = this.props;
-    const sectionBreakImagePath = '/images/section_break_celebration.gif';
+    const { section } = blockShared;
+    const { sections } = this.global;
+    const isFinalBreak = section === sections.length - 2;
```
alexgoff, line 12: since `sectionOrder` is used to create the sections in the global store, it can be used to re-access which section this break is for and in this case we want to see if it is in the next to last section, since the final section is ended by the summary page and QA review

blnkt (approved):

blnkt (approved):
There's something out of order happening with the page navigation for the last handful of pages, will chat with you about it on slack.  Otherwise looks ready to go.

PR #504, opened 2022-12-15
[C] additional reset options

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-7358

## What this change does ##

Add a "Reset" button in addition to the existing "Clear Answers" button that displays when users have a non-fresh local storage, but have not answered any questions

## Notes for reviewers ##

-3 

## Testing ##

Testing can be done by making some change to the investigation, but not answering questions, and returning to the landing page to determine if a "Reset" button is available. Press "Reset" then observe it disappearing and the application returning to a fresh state. Refreshing the page after this should not make it return, neither should changing the page and returning.

Reset should not appear if questions have been answered. 

## Checklist ##

### DIFF
--- a/src/containers/LandingContainer.jsx
+++ b/src/containers/LandingContainer.jsx
@@ -42,6 +42,7 @@ class InvestigationsLanding extends React.PureComponent {
     const { investigations } = pageContext;
     const { envInvestigation } = this.state;
     const isAnswers = Object.keys(this.global.answers).length > 0;
+    const { fromLocal } = this.global;
 
     return (
       <>
@@ -73,7 +74,7 @@ class InvestigationsLanding extends React.PureComponent {
             <div className="space-bottom landing-page-toggle">
               <EducatorModeToggle />
             </div>
-            {isAnswers && (
+            {isAnswers ? (
               <>
                 <br />
                 <br />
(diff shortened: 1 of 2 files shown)


src/containers/LandingContainer.jsx
```diff
@@ -103,6 +104,20 @@ class InvestigationsLanding extends React.PureComponent {
...
+              !fresh && (
+                <div className="space-bottom">
+                  <Button
+                    flat
+                    primary
+                    swapTheming
+                    onClick={this.dispatch.empty}
+                    style={{ backgroundColor: '#df0039' }}
+                  >
+                    <Trans>interface::actions.reset</Trans>
```
alexgoff, line 117: this is just a simple "Reset", but perhaps some language around it would be helpful?

src/state/GlobalStore.js
```diff
@@ -40,17 +40,17 @@ class GlobalStore {
   }
 
   addReducers() {
-    addReducer('empty', () => {
-      resetGlobal();
-      this.addCallbacks();
-      this.addReducers();
+    addReducer('empty', global => {
+      const { investigation } = global;
```
alexgoff, line 44: `resetGlobal()` returns the global state to it's original value, which could be an invalid value loaded from local storage. Instead, delete the local storage and set a fresh empty state.

src/state/GlobalStore.js
```diff
@@ -40,17 +40,17 @@ class GlobalStore {
...
+      remove(investigation);
 
-      return this.emptyState;
+      return { ...this.emptyState, fresh: true };
     });
 
     addReducer('updateName', (global, dispatch, name) => {
       return {
         ...global,
+        fresh: false,
```
alexgoff, line 53: all actions except for updating page ID will make the state no longer fresh.

blnkt (approved):

